PR #2, opened 2024-04-08
chore: setup biome

nhedger:
wip


biome.json
```diff
@@ -0,0 +1,23 @@
+{
+	"$schema": "https://biomejs.dev/schemas/1.6.4/schema.json",
+	"files": {
+		"ignore": ["biome/**"]
```
nhedger, line 4: ignore all files from the submodule

PR #22, opened 2024-04-10
fix: pnpm and netlify scripts

Sec-ant:
## Summary

Fix some pnpm and netlify scripts. We can build the ~~website~~ wasm now.


package.json
```diff
@@ -2,69 +2,72 @@
...
+		"format": "pnpm biome format --write .",
+		"check": "pnpm biome check --apply .",
+		"tsc": "tsc --skipLibCheck",
 		"build": "pnpm build:wasm && pnpm build:js",
 		"build:js": "astro build",
-		"build:wasm": "wasm-pack build --out-dir ../../packages/@biomejs/wasm-web --target web --release --scope biomedev ../crates/biome_wasm --features format_css",
-		"build:wasm-dev": "wasm-pack build --out-dir ../../packages/@biomejs/wasm-web --target web --scope biomedev ../crates/biome_wasm --features format_css",
+		"build:wasm-common": "wasm-pack build --out-dir ../../packages/@biomejs/wasm-web --target web --scope biomedev",
+		"build:wasm": "pnpm build:wasm-common --release ./biome/crates/biome_wasm --features format_css",
+		"build:wasm-dev": "pnpm build:wasm-common --profiling ./biome/crates/biome_wasm --features format_css",
```
Sec-ant, line 16: Here i use the profiling profile because we weren't passing any profile flags before, [which means we've always been using the release profile](https://rustwasm.github.io/docs/wasm-pack/commands/build.html#profile) for dev build.

I initially tried to use the `dev` profile, however that will make `wasm-bindgen` issue an error `locals exceed maximum` like [this one](https://github.com/rustwasm/wasm-pack/issues/981). I tried tweaking the options of the profile in the project root `Cargo.toml` but with no luck. But this error doesn't happen in `profiling` or `release` profiles. So I decided to use `profiling` for dev builds.

The good side of using different profiles is that we can switch off `wasm-opt` for the profiling profile [like this](https://rustwasm.github.io/docs/wasm-pack/cargo-toml-configuration.html), which will save us some time when doing a dev build.

netlify.toml
```diff
@@ -1,3 +1,3 @@
 [build]
 publish = "dist/"
-command = "npm i -g wasm-pack && pnpm run build:wasm && pnpm ci && pnpm run build:js"
+command = "pnpm i --frozen-lockfile && pnpm add -g wasm-pack && pnpm init:biome && pnpm build"
```
Sec-ant, line 3: Put `pnpm i --frozen-lockfile` to the start can make the build script fail fast if the lock file is not updated. So we don't have to waste time on the wasm building process.

`pnpm add -g wasm-pack` should be more cache friendly then `npm i -g wasm-pack` because we use the same package manager. (I don't know how netlify deployment cache works though)

package.json
ematipico: Yeah I hit this issue many times, that's why I eventually used the prod build, which isn't ideal for the website previews.

I did try to set the flags here:

https://github.com/biomejs/biome/blob/dcb695d633d28fc5ee83cbc3cb398acbabd2b9aa/Cargo.toml#L179-L189

But they don't seem to work, probably because the `wasm-pack` CLI ignores them :(

ematipico (approved):

package.json
Sec-ant: Yeah I saw those options. I even applied all the defaults of the release profile to the dev one but the error still occurs. :(

package.json
```diff
@@ -2,69 +2,72 @@
...
 	"scripts": {
+		"init:biome": "git submodule update --init --recursive",
+		"checkout:biome": "cd biome && git checkout",
 		"start": "astro dev",
-		"format": "cargo biome-cli-dev format --write .",
-		"check": "cargo biome-cli-dev check --apply .",
-		"tsc": "tsc --skipLibCheck",
 		"start:playground": "pnpm build:wasm-dev && pnpm start",
+		"format": "pnpm biome format --write .",
+		"check": "pnpm biome check --apply .",
```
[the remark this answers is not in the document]
SuperchupuDev: FYI you can remove the `pnpm` prefix here and it would work too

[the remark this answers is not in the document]
Sec-ant: Thanks! TIL. But I assume `pnpm biome` will always use the local dependency even if the binary can also be found in the system path?

PR #23, opened 2024-04-10
chore: use relative path schema

Sec-ant:
## Summary

So we don't have to manually update when Biome is bumped.

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -1,5 +1,5 @@
 {
-	"$schema": "https://biomejs.dev/schemas/1.6.4/schema.json",
+	"$schema": "./node_modules/@biomejs/biome/configuration_schema.json",
 	"files": {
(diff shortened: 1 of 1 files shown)


ematipico:
We used to use it, although there were a few issues:
- users don't learn about it from our website 
- using relative path has issues with package managers that work with symbolic links e.g. pnpm

Sec-ant:
Good, I'll leave it as is then.

I wonder if we can create a lint rule to check outdated schema? I saw many users still use the schema from earlier versions when they have updated Biome to the latest one.

ematipico:
> Good, I'll leave it as is then.
> 
> I wonder if we can create a lint rule to check outdated schema? I saw many users still use the schema from earlier versions when they have updated Biome to the latest one.

Running `biome migrate --write` will update the schema to the latest

PR #36, opened 2024-04-13
chore(ci): add pin submodule and run codegen workflow

Sec-ant:
## Summary

Add a workflow to bump and pin submodules on `repository_dispatch` event. It can also be triggered by `workflow_dispatch` with a commit id.

After the submodule is updated, the workflow will run codegen to generate potential new rule docs. When everything is done, this workflow will create a PR to update the submodule and the docs.

A newly triggered action will cancel the previous ongoing one of the same trigger event type. And actions triggered by the same event type will land in the same PR head branch.

I also updated the codegen project and cargo scripts. The previous setup has some issues.

`repository_dispatch` issuer is yet to be implemented in the main repo, it will need a personal access token.

Auto merge is implemented for update contributors and pin submodule codegen workflows. However there're some conditions that have to be met:

image


Sec-ant:
I'm going to mark this as draft because due to the nature of github actions I'll have to test them on the main branch. And I'm now testing new things so the main branch will be updated.

ematipico:
I think we have the proper settings now

Sec-ant:
Okay, I'm going to merge this to test whether it'll function in this repo.

PR #66, opened 2024-04-17
chore: update the website content

Sec-ant:
## Summary

I synced the website content with the main repo. If this is not the final transition, this PR still helps because of less diffs and a smoother transition later.


ematipico (approved):
We're almost there! Thank you @Sec-ant

PR #70, opened 2024-04-17
fix: netlify deployment

Sec-ant:
## Summary

This is a (not so) quick fix for the netlify deployment.

1. Exclude nested git submodules to save some time
2. Disable `wasm-opt` and use `pnpm build:wasm-dev` to save some time
3. Update `rust-toolchain.toml`


Sec-ant:
The preview is working now, the total build time is ~5 min. There's definitely some margin to optimize the build time because we haven't started using the netlify cache. I'll play aroung with different options that mentioned in the discord server later.

ematipico (approved):
Oh yeah... submodules of submodules 😅

PR #75, opened 2024-04-18
Update number of lint rules

chansuke:
## Summary

While reading the Japanese documentation today, I noticed that the numbers of lint rules in each README are outdated.
I have updated them accordingly.


Sec-ant:
Thank you!

PR #84, opened 2024-04-19
chore: cache WASM artifacts

Sec-ant:
## Summary

Use [Netlify build plugins](https://docs.netlify.com/integrations/build-plugins/) to cache WASM artifacts. Deployments with the same submodule commit ID will reuse the cached WASM artifacts. Cached artifacts will be preserved for ~30 days.

Skipping the WASM building process will save ~1 min of build time and reduce the build time to ~3min. To further optimize the build time we need to focus on the astro building process.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,6 +1,6 @@
 [build]
 publish = "dist/"
-command = "pnpm i --frozen-lockfile && npm i -g wasm-pack && pnpm build:wasm-dev && pnpm build:js"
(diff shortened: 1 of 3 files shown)


Sec-ant:
WASM building process can be skipped now if the artifacts corresponding to that submodule commit ID are cached earlier:

image

ematipico (approved):
Great documentation, thank you!


Sheep it

PR #142, opened 2024-04-25
refactor: improve linter and formatter intros

Conaclos:
## Summary

This PR improves the linter and the formatter introductions.

- I removed the list of recommended rules from the linter introduction.
  This was too verbose and this information can be found in the `rules` page.
- I added an `Ignoring files` section to the two introductions.
- I refactored the structure of the linter intro, hopefully making it more readable
- I added details about how rules are organized, their severity level and more

I still have some ideas to improve the website.
However, it is already a large chunk of changes.


ematipico:
>  This was too verbose and this information can be found in the rules page.

True, it's verbose, however, they aren't grouped together, so it's difficult to see all of them together :/

Conaclos:
> True, it's verbose, however, they aren't grouped together, so it's difficult to see all of them together :/

It would be nice to have filters on the rules page, like [TypeScript ESLint](https://typescript-eslint.io/rules/) does.
In the meantime we could add the list of recommended rules on the `rules` page?

ematipico:
Yep! Sounds good

src/content/docs/formatter/index.mdx
```diff
@@ -5,36 +5,44 @@ description: How to use the Biome formatter.
...
 
 If you want to **apply** the new formatting, pass the `--write` option:
 
 <PackageManagerBiomeCommand command="format --write ./src" />
 
-Use the  `--help` flag to know what are the available options:
+The command accept a list of files and diretcories.
+Note that if you pass a glob as a parameter, it will be expanded by your shell.
+The result of the expansion depends on your shell and its configuration.
+For example, some shells don't support the recursive glob `**` or the alternation `{}` in the following command:
```
ematipico, line 28:
```suggestion
The command accepts a list of files and directories.
Note that if you pass a glob as a parameter, your shell will expand it.
The result of the expansion depends on your shell and its configuration.
For example, some shells don't support the recursive glob `**` or the alternation `{}` in the following command:
```

Maybe we should put this paragraph inside a `:::note`, it's something require a bit of attention

ematipico: Also, we should add a `:::caution` that says that shell expansion of patterns might slow down the command by a lot.

src/content/docs/formatter/index.mdx
```diff
@@ -111,3 +119,30 @@ const expr =
...
+## Ignoring files
+
+Biome doesn't provide ignore comments that ignore an entire file.
+However, you can ignore a file using the Biome configuration file.
+
+You can ignore files for all tools, including the formatter,
+using the [`files.ignore` configuration](/guides/how-biome-works#include-and-ignore-explained).
+By default, Biome ignores the [protected files](/guides/how-biome-works#protected-files).
+
+If you want that only the formatter to ignore some files, you can use `linter.ignore`:
```
ematipico, line 133:
```suggestion
If you want to exclude files from being formatted, you can use `formatter.ignore`:
```

src/content/docs/linter/index.mdx
```diff
@@ -4,180 +4,205 @@ description: How to use the Biome linter.
 ---
 
 import NumberOfRules from "@/components/generated/NumberOfRules.astro";
-import RecommendedRules from "@/components/generated/RecommendedRules.astro";
+import PackageManagerBiomeCommand from "@/components/PackageManagerBiomeCommand.astro";
 
-Biome's linter statically analyzes your code to catch common errors and to help writing idiomatic code.
+Biome's linter statically analyzes your code to find and fix common errors and to help you write better code.
```
ematipico, line 9:
```suggestion
Biome's linter statically analyzes your code to find and fix common errors and to help you write better, modern code.
```

src/content/docs/linter/index.mdx
```diff
@@ -4,180 +4,205 @@ description: How to use the Biome linter.
...
 
-## Rule pillars
-
-We believe that rules should be informative and explain to the user why a rule is triggered and tell the user what they should to do fix the error. A rule should follow these **pillars**:
-1. Explain to the user the error. Generally, this is the message of the diagnostic.
-2. Explain to the user **why** the error is triggered. Generally, this is implemented with an additional node.
-3. Tell the user what they should do. Generally, this is implemented using a code action. If a code action is not applicable a note should tell the user what they should do to fix the error.
+Note that if you pass a glob as a parameter, it will be expanded by your shell.
+The result of the expansion depends on your shell and its configuration.
+For example, some shells don't support the recursive glob `**` or the alternation `{}` in the following command:
```
ematipico, line 21:
```suggestion
Note that if you pass a glob as a parameter, your shell will expand it.
The result of the expansion depends on your shell and its configuration.
For example, some shells don't support the recursive glob `**` or the alternation `{}` in the following command:
```

We have the same phrase in the formatter. Maybe we should have it in a common section, otherwise we will struggle to update both.

src/content/docs/linter/index.mdx
```diff
@@ -4,180 +4,205 @@ description: How to use the Biome linter.
...
 
-Lint rules may provide automatic code fixes. Biome distinguishes between two types of fixes.
 
-### Safe fixes
+## Rules
 
-Safe fixes are guaranteed to not change the semantic of your code. They can be applied without explicit review.
+The linter is organized into rules.
+A rule emit a diagnostic when it encounters a code that doesn't meet its requirements.
+For example, the [noDebugger](http://localhost:4321/linter/rules/no-debugger) rule reports the use of the `debugger` instruction in JavaScript code.
```
ematipico, line 34:
```suggestion
A rule emits a diagnostic when it encounters a code that doesn't meet its requirements.
For example, the [noDebugger](http://localhost:4321/linter/rules/no-debugger) rule reports using the `debugger` instruction in JavaScript code.
```

src/content/docs/linter/index.mdx
```diff
@@ -4,180 +4,205 @@ description: How to use the Biome linter.
...
 
-To apply _safe fixes_, use `--apply`:
+A rule emits diagnostics with a `warn` or `error` severity.
+Diagnostics with an `error` severity cause the command to exit with a non-zero code,
+While diagnostics with a `warn` severity don't cause the command to fail.
+You can turn warnings into errors `--error-on-warnings` CLI option:
```
ematipico, line 39: I would change the explanation and focus more on the rules rather than discussing diagnostics. In the future, a full, in-depth explanation of our diagnostics should be on a separate page.

My suggestion:

```suggestion
By default, **recommended** rules emit a diagnostic with `error` severity and will cause the command to fail. Via configuration, you can demote recommended rules to emit warnings or promote non-recommended rules to emit errors.

You can turn warnings into errors `--error-on-warnings` CLI option:
```

src/content/docs/linter/index.mdx
```diff
@@ -4,180 +4,205 @@ description: How to use the Biome linter.
...
 
-### Unsafe fixes
+By default, the Biome linter only runs [**recommended rules**](/linter/rules/).
+Recommended rules emit diagnostics with the `error` severity.
 
-Unsafe fixes may change the semantic of your program. Therefore, it's advised to manually review the changes.
+Rules are organized into groups.
+For example, the `noDebugger` rule is part of the [`suspicious` group](/linter/rules#suspicious).
+A Rule from this group detects code that is likely to be incorrect or useless.
+The description of each group can be found on the [rules page](/linter/rules/).
```
ematipico, line 51:
```suggestion
Our [rules page](/linter/rules/) explains each group in-depth.
```

There is no need to explain a group if there's a page already for that (we avoid repeating information and causing too much fragmentation).

src/content/docs/linter/index.mdx
```diff
@@ -4,180 +4,205 @@ description: How to use the Biome linter.
...
 
-To apply _unsafe fixes_, use `--apply-unsafe`:
+Unlike other linters, we don't provide any rules that check for code formatting.
+This kind of checking is covered by our [code formatter](/formatter/).
 
-```shell
-biome check --apply-unsafe ./src
-```
+Many rules provide a **code fix** that can be automatically applied.
+Biome distinguishes between safe and unsafe code fixes.
```
ematipico, line 57:
```suggestion
Biome distinguishes between **safe** and **unsafe** code fixes.
```

src/content/docs/linter/index.mdx
```diff
@@ -4,180 +4,205 @@ description: How to use the Biome linter.
...
-// biome-ignore lint: reason
-debugger;
-// biome-ignore lint/suspicious/noDebugger: reason
-debugger;
-```
 
 ## Configuration
 
-### Enable a lint rule
+Recommended rules are enabled by default and emit diagnostics with the `error` severity.
```
ematipico, line 95:
```suggestion
```

Repeated phrase

src/content/docs/linter/index.mdx
```diff
@@ -4,180 +4,205 @@ description: How to use the Biome linter.
...
+Rules that are not recommended are disabled by default, but they can be enabled via [configuration](/guides/how-biome-works#configuration).
```
ematipico, line 96: If you accept my previous suggestion, this is also a repetition.

src/content/docs/linter/index.mdx
```diff
@@ -4,180 +4,205 @@ description: How to use the Biome linter.
...
 
-Recommended rules are enabled by default and emit diagnostics with the error severity.
-Rules that are not recommended are disabled by default, but they can be enabled via configuration.
-The diagnostics emitted by these rules are displayed with the warning severity in the documentation.
+A rule is enabled whether its severity is either `error` or `warn` and is disabled if its severity is `off`.
```
ematipico, line 98: In our configuration, we have `level`, but here you say, "Its severity is off." I believe this is confusing. I think we shouldn't say "severity", and just say "you can turn off a rule with `"off"`"

src/content/docs/linter/index.mdx
```diff
@@ -4,180 +4,205 @@ description: How to use the Biome linter.
...
+## Ignoring files
+
+Biome doesn't provide ignore comments that ignore an entire file.
+However, you can ignore a file using the Biome configuration file.
+
+You can ignore files for all tools, including the linter,
+using the [`files.ignore` configuration](/guides/how-biome-works#include-and-ignore-explained).
+By default, Biome ignores the [protected files](/guides/how-biome-works#protected-files).
+
+If you want that only the linter to ignore some files, you can use `linter.ignore`:
```
ematipico, line 184: A lot of repeated words equals the formatter. I am worried that it could take too much work if a tomorrow will change things, for example, around how we ignore files. 

We should try to keep things in one place, and use links to redirect users.

ematipico (commented):
Overall, it's a better improvement! I left some suggestions to avoid repeating the exact phrases more than once.

src/content/docs/linter/index.mdx
Conaclos: I will try to address the duplication concern in a future PR.

src/content/docs/linter/index.mdx
Conaclos: I will try to address the duplication concern in a future PR.

Conaclos:
I added some badges as you suggested @ematipico.

ematipico (approved):

PR #140, opened 2024-04-24
docs: add renovate recipe

setchy:
## Summary

Closes #122

Welcome any edits or feedback 🙏

PR #150, opened 2024-04-27
docs: use valid JSON in Renovate config example

HonkingGoose:
## Summary

Follow up PR after:

- #140

The config example should use valid JSON. The `...` characters are not allowed in JSON files.

Having valid JSON also means that readers can copy/paste the JSON into their `renovate.json` file.

**Edit:** I'm also dropping the word `simply` from the docs. If you need to explain something, it's not simple. Also shorter sentences are easier to read and remember.

> [!TIP]
> If you're copy/pasting the example into your `renovate.json` file, then also add `config:recommended` _or_ `config:best-practices` to the `extends` array. This way Renovate will update your other dependencies as well. But that's a bit off topic for the Biome recipe example, so I'm not putting it in the docs here.


nhedger (approved):

nhedger:
Thanks!

HonkingGoose:
Thank you for the quick review and merge! :heart:

PR #174, opened 2024-04-29
refactor: tldr migratation guide & noUnusedImports ref in import sorting

Conaclos:
## Summary

- Add a reference to `noUnusedImports` on the import sorting page
- Add a TLDR on the migration guide

### DIFF
--- a/codegen/Cargo.lock
+++ b/codegen/Cargo.lock
@@ -661,7 +661,6 @@ dependencies = [
  "rustc-hash",
  "serde",
  "serde_json",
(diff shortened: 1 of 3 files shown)


biome
Sec-ant, line 1: I believe this is an accidental commit. I think we need some mechanism to prevent developers from accidentally mutate this submodule ref. This should only be modified by the github actions.

Conaclos: Thanks! Yeah this is pretty annoying to check!

PR #214, opened 2024-05-02
chore(ci): auto codegen release files

Sec-ant:
## Summary

This PR improves the pin submodule and run codegen workflow.

Previously, when we do a release commit in the main repo, we'll also have to release the files in this website repo to generate the changelog, schema url, etc. This PR automates the releasing files generation.

When the commit message of the main repo commit is in the form of `release: vX.X.X`, the codegen workflow will run `cargo run --manifest-path codegen/Cargo.toml all` with the environment variable `BIOME_VERSION` set to the parsed version from that commit message.

I was planning to use the same version check github action [EndBug/version-check](https://github.com/EndBug/version-check) that is used in the main repo to detect new versions, but that action doesn't fit our submodule and repository dispatch setup, so I used a simple bash script and `sed` to extract the version. Therefore it is very ad hoc and the release commit in the main repo has to follow the pattern strictly.

I think this should be aware of by all the core contributors, @biomejs/core-contributors.

## Test

This workflow is tested by manually triggering it with a release commit id.

image


arendjr (approved):

ematipico:
@Sec-ant I don't think checking the commit message is future proof, especially if we consider that in the near future we will automate how we do releases. Since we have a monorepo, with the automation tool that we chose, we might not be able to create a PR that has the version number in the title.

Sec-ant:
@ematipico OK, then I will mark this PR as a draft and revisit it once we migrate to the new release workflow.

Sec-ant:
> Can't we just have a `workflow_dispatch` workflow that we trigger manually, and it pushes the new changes to production? We can add this step to the list of things to do when doing a release.

I can add another input for "pin submodule and run codegen" to indicate the type of codegen we want to run and also add an input to get the version number.

I just think the whole release process should be atomic and uninterruptable once it is triggered in the main repo. So an automated run should be the default behavior.

ematipico:
> I just think the whole release process should be atomic and uninterruptable once it is triggered in the main repo. So an automated run should be the default behavior.

Oh yeah, I completely agree with you. I am sure we will able to land the perfect solution for us, gradually

Sec-ant:
Many have changed since. Close this for now.

Issue #53, opened 2024-04-16
📎 Prettier in the Playground recognizes JSON as JavaScript


Conaclos:
### Description

In the Playground we can add a JSON file. Biome correctly recognized it as JSON and format this as a JSON file. Prettier seems to recognize the file as JavaScript, because it removes the quotes around the property.

See this [Playground example](https://biomejs.dev/playground/?files.test.json=ewAKACAAIAAiAHAAcgBvAHAAIgA6ACAAdAByAHUAZQAKAH0A).

The Playground should correctly inform Prettier than the file is a JSON file.

Conaclos:
Note: I had to create the issue on the main repo, because the website repo has a single issue template for docs

nhedger:
> Note: I had to create the issue on the main repo, because the website repo has a single issue template for docs

You can open arbitrary issues with this link
<img width="1227" alt="Screenshot 2024-04-17 at 18 35 09" src="https://github.com/biomejs/website/assets/649677/618bf87b-d2af-403d-876c-ad7fed3ad8e2">

PR #228, opened 2024-05-03
fix(playground): use correct JSON parser in prettier worker

ematipico:
## Summary

Closes https://github.com/biomejs/website/issues/53


ematipico:
Checked the plyaground. Fixed!

PR #268, opened 2024-05-08
Fix typo

CHE1RON:
Title says it all 😅


Sec-ant:
Thank you! I think this is a generated file so would you mind also fixing the source here? https://github.com/biomejs/website/blob/de9bb0da69884bbbee103ec9507a20528851f9ed/codegen/src/lintdoc.rs#L223

CHE1RON:
✅

PR #430, opened 2024-05-21
docs: adding info on running Biome via ALE in Vim & Neovim.

lanker:


ematipico:
Is there a direct link to how to set up Biome?

ematipico:
I was just curious, because other LSPs required some custom confirmation

PR #435, opened 2024-05-22
chore: remove git module in favour of cargo git

ematipico:
## Summary

This PR does the following things:
- Removes the use of git modules from the repository.
- Use cargo git to retrieve the dependencies instead of relying on the git module.
- `rust-toolchain.toml` is now a real file, that rarely updates and it won't be an issue.
- Adds a top-level `Cargo.toml`. This is useful with IDEs, where if there isn't a top-level manifest, people were forced to open `codege/` as root folder. It's just better DX, nothing else.
- Copies the code from https://github.com/biomejs/biome/pull/1028/files, to continue the works to make the configuration page autogenerated. I committed `configuration_v2.mdx` to as a starting point.


Sec-ant:
> @Sec-ant, can you check if this commit makes sense? [`e6dee2e` (#435)](https://github.com/biomejs/website/pull/435/commits/e6dee2e7de209e202c53f92b88fbad2617a92a0e)

I'd say it should work but should we specify [the revision of the git dependencies in `Cargo.toml`](https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html)? Otherwise the commit ID we acquired from the action is useless. And there might be some race conditions that can make problems harder to debug.

Although, I'm not sure what's the proper way to update git repo dependencies revisions, do we have to edit the toml file by hand or if there're some convenient commands to do that?

Sec-ant:
We should also temporarily disable the netlify build plugin, becasue it was meant to work with the submodule.

Remove this one:
https://github.com/biomejs/website/blob/6d89c2f3a8dbdfc7ea3c6c1594ddabf8754ac6d4/netlify.toml#L31-L32

And change the build command to: `pnpm i --frozen-lockfile && npm i -g wasm-pack && pnpm build:wasm-dev && pnpm build:js`
https://github.com/biomejs/website/blob/6d89c2f3a8dbdfc7ea3c6c1594ddabf8754ac6d4/netlify.toml#L3-L9

ematipico:
> > @Sec-ant, can you check if this commit makes sense? [`e6dee2e` (#435)](https://github.com/biomejs/website/pull/435/commits/e6dee2e7de209e202c53f92b88fbad2617a92a0e)
> 
> I'd say it should work but should we specify [the revision of the git dependencies in `Cargo.toml`](https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html)? Otherwise the commit ID we acquired from the action is useless. And there might be some race conditions that can make problems harder to debug or reproduce.
> 
> Although, I'm not sure what's the proper way to update git repo dependencies revisions, do we have to edit the toml file by hand or if there're some convenient commands to do that?

What do you think of this change? [`c1dd477` (#435)](https://github.com/biomejs/website/pull/435/commits/c1dd477a5346592e9218b84c42c65a015aec4fb9)

codegen/Cargo.toml
```diff
@@ -7,25 +7,25 @@ publish     = false
...
-biome_json_analyze   = { path = "../biome/crates/biome_json_analyze" }
-biome_json_formatter = { path = "../biome/crates/biome_json_formatter" }
-biome_json_parser    = { path = "../biome/crates/biome_json_parser" }
-biome_json_syntax    = { path = "../biome/crates/biome_json_syntax" }
-biome_rowan          = { path = "../biome/crates/biome_rowan" }
-biome_service        = { path = "../biome/crates/biome_service" }
-biome_string_case    = { path = "../biome/crates/biome_string_case" }
+biome_analyze        = { git = "https://github.com/biomejs/biome.git" }
+biome_cli            = { git = "https://github.com/biomejs/biome.git" }
+biome_configuration   = { git = "https://github.com/biomejs/biome.git" }
```
Sec-ant, line 12: Couldn't help to remove the space 😅

```suggestion
biome_configuration  = { git = "https://github.com/biomejs/biome.git" }
```

Sec-ant:
> What do you think of this change? [`dd1a028` (#435)](https://github.com/biomejs/website/pull/435/commits/dd1a028174b1beabab27f881c9133f028fac2ccf)

Very nice, LGTM

But we still have another loose end: How should we build the WASM file for the playground?:

https://github.com/biomejs/website/blob/7d6d430107ac9c4e6a30ef283637922069ebe4c8/package.json#L33

https://github.com/biomejs/website/blob/7d6d430107ac9c4e6a30ef283637922069ebe4c8/package.json#L14-L16

ematipico:
We can use git dependencies: https://docs.npmjs.com/cli/v10/configuring-npm/package-json#git-urls-as-dependencies

I am sure we can come up with the correct combination

Sec-ant:
> We can use git dependencies: https://docs.npmjs.com/cli/v10/configuring-npm/package-json#git-urls-as-dependencies
> 
> I am sure we can come up with the correct combination 

In the current workflow, we use `wasm-pack` to build the WASM binary inside the submodule dir, and we the use that local filepath for package resolution.

If we use git dependencies for package resolution, that means the WASM binary should be already existed in that remote repo 🤔

Sec-ant:
How about setting up a workflow to continously build `WASM` files/packages in the main repository? This was suggested by @nhedger earlier, and we also have a promising solution with this tool https://github.com/stackblitz-labs/pkg.pr.new.

I just think that all the patches and hassles are beginning to outweigh the benefits of removing the submodule.

ematipico:
Yeah, that's a great idea

PR #449, opened 2024-05-23
docs: update the cli document for --write and --fix options

### LINKED ISSUE #398: 📚 update for `--write`, `--fix`, `--unsafe` options and deprecation of `--apply` and `--apply-unsafe`
### Documentation URL

https://biomejs.dev/reference/cli

### Description

I created PRs
- to add `--write`, `--fix`, `--unsafe` options to fixable commands.
- to deprecate `--apply` and `--apply-unsafe` options of `biome <check|lint>`

The website documentation should be updated for the change.

### Expectations

The page should reflect the actual command options.

### Code of Conduct

unvalley:
## Summary

Closes #398

### DIFF
--- a/src/content/docs/reference/cli.mdx
+++ b/src/content/docs/reference/cli.mdx
@@ -202,7 +202,7 @@ Stop the Biome daemon server process
 
 Runs formatter, linter and import sorting to the requested files.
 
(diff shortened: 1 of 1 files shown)


ematipico:
We can create a label for these cases. Something like "next release". What do you think?

unvalley:
@ematipico 
yeah, I'll do. 
I think this kind of update will be necessary in the future, and there are some ways to do it, such as separating branches for next release and managing them in different PRs, so it might be something to consider.

ematipico:
Ah, we already have that. It's done automatically at each release

ematipico:
What you want to update are the other pages of the website that still use `--apply`. The reference CLI is code generated, and we have a workflow job that does that for us.

PR #450, opened 2024-05-23
docs: fix typo

jer3m01:
## Summary
I'm pretty sure this is a typo as the input file above has different content than the diff:

image


Sec-ant (approved):
Thank you!

PR #452, opened 2024-05-24
docs: Fix typo

jeremyhon:
Use backticks correctly

## Summary

There's two typos in the paragraph where a single quote is used instead of a backtick.


Sec-ant:
This is some [old-school quoting style](https://stackoverflow.com/a/78287468) but apparently it's breaking the markdown. So thank you for the fix!

PR #513, opened 2024-06-05
docs: delete `css.formatter.indentSize` param

Marukome0743:
## Summary
`css.formatter.indentSize` was renamed `css.formatter.indentWidth` at https://github.com/biomejs/biome/pull/2790.

Unlike the other deprecated indentSize options (e.g. [formatter.indentSize](https://biomejs.dev/reference/configuration/#formatterindentsize)),
`css.formatter.indentSize` doesn't work anymore.

That's why `css.formatter.indentSize` should be deleted.

### DIFF
--- a/src/content/docs/reference/configuration.mdx
+++ b/src/content/docs/reference/configuration.mdx
@@ -769,12 +769,6 @@ The style of the indentation for CSS (and its super languages) files. It can be 
 > Default: `"tab"`
 
 
-### `css.formatter.indentSize`
-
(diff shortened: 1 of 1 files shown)


Sec-ant (approved):
Oh yes I believe this is a leftover. CSS formatter wasn't a thing before v1.8 and there's no point introducing a deprecated option in a new feature. That's why `indentSize` isn't a deprecated alias for the CSS formatter like other ones.

Thank you!

PR #518, opened 2024-06-06
feat: auto-generate configuration from source code

ematipico:
## Summary

This is my attempt to create the configuration reference page from source code.

The drawback: we lose many examples
The con: we keep one source of truth

I see two options here:
- Making our doc comments in the source more rich, with markdown examples and more. They will be rendered here.
- Having use cases and examples in a separate page.

What do you think?

### DIFF
--- a/codegen/src/configuration.rs
+++ b/codegen/src/configuration.rs
@@ -1,36 +1,126 @@
 //! Auto generate markdown documentation for the configuration schema.
+use biome_configuration::PartialConfiguration;
+use schemars::gen::{SchemaGenerator, SchemaSettings};
+use schemars::schema::{RootSchema, Schema, SchemaObject};
+use std::collections::VecDeque;
 use std::fs::File;
 use std::io::Write;
-use biome_configuration::PartialConfiguration;
 
 use schemars::schema_for;
 use serde_json::{to_string_pretty, Value};
 
-pub fn generate_configuration() -> anyhow::Result<()> {
-    let schema = schema_for!(PartialConfiguration);
-    let json_schema = to_string_pretty(&schema)?;
-    let schema: Value = serde_json::from_str(&json_schema)?;
-    let mut markdown = File::create("src/content/docs/reference/configuration_v2.mdx")?;
+struct Queue<'a> {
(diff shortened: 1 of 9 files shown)


ematipico:
https://deploy-preview-518--biomejs.netlify.app/reference/configuration_v2/

Sec-ant:
> The drawback: we lose many examples
> The con: we keep one source of truth

Aren't "drawback" and "con" the same thing? 😄

I would prefer keeping the options and examples together so I don't have to jump back and forth across pages. And the preview link further determined my thought.

So it seems richer markdown doc comments the way to go. But I get the impression that `pulldown-cmark` doesn't handle rich markups very well. 🤔

ematipico:
> Aren't "drawback" and "con" the same thing? 😄

That's how damanged my brain is 😆 

> I would prefer keeping the options and examples together so I don't have to jump back and forth across pages. And the preview link further determined my thought.

Yeah, that's what the [original PR](https://github.com/biomejs/biome/pull/1028) attempts to do, and I don't mind moving better docs and examples in the source code.

> So it seems richer markdown doc comments the way to go. But I get the impression that pulldown-cmark doesn't handle rich markups very well. 🤔

I don't think we need `pulldown-cmark` for now. I think we just need to take the doc comments as they are, and write them.

codegen/src/configuration.rs
```diff
@@ -0,0 +1,350 @@
...
+    let mut buffer = Vec::new();
+
+    generate_buffer(&mut buffer)?;
+
+    fs::write("src/content/docs/reference/configuration_v2.mdx", buffer)?;
+
+    Ok(())
+}
+
+fn generate_markdown_hearer() -> String {
```
chansuke, line 255: i noticed a typo in my original [function name](https://github.com/biomejs/biome/pull/1028/files#diff-bfd7157fe097ef23de66a65587759406e2ad7105b057b3915d097feb3565bd49R10)... :pray:  It should be `generate_markdown_header()` and sorry for the mistake.

Sec-ant:
> I don't think we need pulldown-cmark for now. I think we just need to take the doc comments as they are, and write them.

Oh yes, I was wrong.

PR #527, opened 2024-06-07
docs: add `javascript.formatter.trailingCommas`

Marukome0743:
## Summary
The `javascript.formatter.trailingComma` option is deprecated and renamed to `javascript.formatter.trailingCommas`.
Then add the configuration.

Details can be checked in https://github.com/biomejs/biome/pull/2492

### DIFF
--- a/src/content/docs/reference/configuration.mdx
+++ b/src/content/docs/reference/configuration.mdx
@@ -527,6 +527,15 @@ Print trailing commas wherever possible in multi-line comma-separated syntactic 
 
 > Default: `"all"`
 
+### `javascript.formatter.trailingCommas`
+
+Print trailing commas wherever possible in multi-line comma-separated syntactic structures. Possible values:
+- `"all"`, the trailing comma is always added;
+- `"es5"`, the trailing comma is added only in places where it's supported by older version of JavaScript;
+- `"none"`, trailing commas are never added.
+
+> Default: `"all"`
+
 ### `javascript.formatter.semicolons`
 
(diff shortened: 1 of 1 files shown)


Sec-ant (approved):
Thank you, I think we should also mark `trailingComma` as deprecated or remove it. Do you want to help?

There're other deprecated options that possibly should also be taken care of, like `indentSize` and command line options like `--trailing-comma` in the CLI doc page.

Marukome0743:
> Thank you, I think we should also mark `trailingComma` as deprecated or remove it. Do you want to help?

Yup, I wanna add the warning to old options.

> [!WARNING]
> `javascript.formatter.trailingComma` is deprecated.
> Use `javascript.formatter.trailingCommas` instead.

Or folding the old options like the below.

<details>
<summary>javascript.formatter.trailingComma (deprecated)</summary>

Print trailing commas wherever possible in multi-line comma-separated syntactic structures. Possible values:

- `"all"`, the trailing comma is always added;
- `"es5"`, the trailing comma is added only in places where it’s supported by older version of JavaScript;
- `"none"`, trailing commas are never added.
Default: `"all"`
</details>

---

> There're other deprecated options that possibly should also be taken care of, like `indentSize` and command line options like `--trailing-comma` in the CLI doc page.

I'm willing to update the docs in order to adapt version 1.8.0.
However, I wonder my changes will be overriden when #518 merged.

This PR might be just temporary update until #518.

Sec-ant:
Fair, let's merge this first.

PR #541, opened 2024-06-08
chore: update biome configuration and scripts

ematipico:
## Summary
The `biome/**` was causing weird issue on my machine. After I deleted it, the scripts started to work again

Closes #533

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
 .astro/
 dist
-
+node_modules
 target/
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,4 @@
 .astro/
 dist
-node_modules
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -1,6 +1,4 @@
 .astro/
 dist
-node_modules
```
Sec-ant, line 3: We shouldn't remove `node_modules`.

ematipico: Why? Biome ignores it already

Sec-ant: But this is for `git` to ignore it, isn't it?

ematipico: Lol true 😅

Sec-ant (approved):
This closes #533

PR #591, opened 2024-06-12
Remove unnecessary example

lanker:
The two examples given only differed on how the div element was closed (self-closing vs non-self-closing), which is not relevant to the rule at hand. To avoid confusing the reader, let's keep only one of the examples.

### DIFF
--- a/src/content/docs/linter/rules/use-key-with-click-events.md
+++ b/src/content/docs/linter/rules/use-key-with-click-events.md
@@ -22,22 +22,6 @@ This does not apply for interactive or hidden elements.
 ### Invalid
 
 ```jsx
-<div onClick={() => {}} />
-```
-
-<pre class="language-text"><code class="language-text">code-block.jsx:1:1 <a href="https://biomejs.dev/linter/rules/use-key-with-click-events">lint/a11y/useKeyWithClickEvents</a> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
-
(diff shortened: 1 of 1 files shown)


Sec-ant:
Hi @lanker

- The page is generated from the source code. So to remove the example, you'll have to edit the [document comments in the source](https://github.com/biomejs/biome/blob/0d9b60a181cf59004815c928c277d13f60fb3b9b/crates/biome_js_analyze/src/lint/a11y/use_key_with_click_events.rs#L16-L19).
- I think both of the two examples exist because we want to make sure the rule works on both the openning element and the self-closing element (We run our analyzer on the code blocks in this page when building it). But I agree this may confuse the reader and I also noticed we already have [dedicated test cases](https://github.com/biomejs/biome/blob/0d9b60a181cf59004815c928c277d13f60fb3b9b/crates/biome_js_analyze/tests/specs/a11y/useKeyWithClickEvents/invalid.jsx#L3-L4) for this purpose. So I think removing one of them is ok.
- Due to the rule [useSelfClosingElements](https://biomejs.dev/linter/rules/use-self-closing-elements/) exists, what do you think of removing the other example, simply for consistency?

lanker:
> Hi @lanker

Hi, thanks for the quick feedback!

>     * The page is generated from the source code. So to remove the example, you'll have to edit the [document comments in the source](https://github.com/biomejs/biome/blob/0d9b60a181cf59004815c928c277d13f60fb3b9b/crates/biome_js_analyze/src/lint/a11y/use_key_with_click_events.rs#L16-L19).

Ah! Sorry for not looking more carefully how it's done.

>     * I think both of the two examples exist because we want to make sure the rule works on both the openning tag and the self-closing element (We run our analyzer on the code blocks in this page when building it). But I agree this may confuse the reader and I also noticed we already have [dedicated test cases](https://github.com/biomejs/biome/blob/0d9b60a181cf59004815c928c277d13f60fb3b9b/crates/biome_js_analyze/tests/specs/a11y/useKeyWithClickEvents/invalid.jsx#L3-L4) for this purpose. So I think removing one of them is ok.
> 
>     * Given that the rule [useSelfClosingElements](https://biomejs.dev/linter/rules/use-self-closing-elements/) exists, what do you think of removing the other example, simply for consistency?

Yep, makes sense to keep the self-closing example instead. I opened a new PR towards the biome repo: https://github.com/biomejs/biome/pull/3185 and will close this one.

PR #604, opened 2024-06-14
fix: enable property type error in config example

quanru:
## Summary

The enable property type is wrong in config example.


Sec-ant (approved):
Thanks!

PR #609, opened 2024-06-14
fix: syntax highlighting for cjs code blocks

Sec-ant:
## Summary

Shiki doesn't recognize `cjs` code blocks, it will fallback to render them as `plaintext`.

For example: https://biomejs.dev/linter/rules/no-redundant-use-strict/

Also check the deployment log: https://app.netlify.com/sites/biomejs/deploys/666c5e50c6f7750008bb8102#L143-L156

I patched the astro config to load the js grammar for `cjs` codeblocks.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+shamefully-hoist=true
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,3 @@
+# Hoist the shiki package so we can import language grammars from it.
+# We use this approach to add syntax highlighting support for the cjs code block.
 shamefully-hoist=true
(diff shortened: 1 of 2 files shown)


.npmrc
```diff
@@ -0,0 +1 @@
+shamefully-hoist=true
```
ematipico, line 1: It might be worth documenting why we need it, just in case, so we don't forget it's because of shiki patching

ematipico (approved):
Should we file a bug/feature request to shiki?

.npmrc
Sec-ant: You're right. Done.

Sec-ant:
> Should we file a bug/feature request to shiki?

Will do my tomorrow

Sec-ant:
Actually there is a [`langAlias`](https://shiki.matsu.io/guide/load-lang#custom-language-aliases) option in shiki that allows registering aliases for a language, but it is not exposed from `astro` nor `starlight`. Not sure if astro will consider supporting it?

ematipico:
> Not sure if astro will consider supporting it?

I think they would

Sec-ant:
https://github.com/withastro/roadmap/discussions/954

PR #640, opened 2024-06-20
feat: support for graphql files

ematipico:
## Summary

This PR updates the frontend to support `.graphql` and `.gql` files, however the playground is throwing an error.

This is probably a bug  in core, I will look after it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -78,7 +78,7 @@ dependencies = [
 [[package]]
 name = "biome_analyze"
 version = "0.5.7"
-source = "git+https://github.com/biomejs/biome.git?rev=54b4c9a39078836843ef363bb6986fad74834480#54b4c9a39078836843ef363bb6986fad74834480"
+source = "git+https://github.com/biomejs/biome.git?rev=ad7881a16b38ea930fa668bce5196cae38bef3de#ad7881a16b38ea930fa668bce5196cae38bef3de"
 dependencies = [
  "biome_console",
  "biome_deserialize",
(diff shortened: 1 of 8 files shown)


Sec-ant:
We need to exlude graphql for the control flow graph: https://github.com/biomejs/website/blob/555dee4a30bdb234ac4c35349564d930f711bfcd/src/playground/workers/biomeWorker.ts#L222-L229

Sec-ant:
I just realized we might have to exlude graphql from all the capabilities including formatting and linting, not only control flow graph. Otherwise those `workspace.` functions will throw errors:

https://github.com/biomejs/website/blob/555dee4a30bdb234ac4c35349564d930f711bfcd/src/playground/workers/biomeWorker.ts#L231-L261

ematipico:
True, but would consider this an optimisation, not a blocker

Sec-ant:
> True, but would consider this an optimisation, not a blocker

Yeah, but this problem prevents us from [using graphql code in the playground](https://deploy-preview-640--biomejs.netlify.app/playground/?files.main.gql=cQB1AGUAcgB5ACAAewAgAAoAIAAgAHYAaQBlAHcAZQByACAAewAgAAoAIAAgACAAIABsAG8AZwBpAG4ACgAgACAAfQAKAH0A), no?

ematipico:
Good call on this one. I applied your suggestions and now we are getting the correct CST :)

ematipico:
I am going to merge this, then land https://github.com/biomejs/biome/pull/3248 and then I will make another PR here to enable `pullDiagnostics`. That should give us a good output for the diagnostics.

PR #644, opened 2024-06-21
fix: graphql format output type

Sec-ant:
## Summary

Fix a type issue that may break the playground.


ematipico (approved):

ematipico:
I was looking at that and wondered why the type is `any`.

We have two options:
- we make the `workspace_types.rs` better
- we change the types returned by those functions to match what returned by the other functions

PR #648, opened 2024-06-21
docs: Fix use of deprecated `trailingCommas` in biome.json example

frodeaa:
## Summary

Update the example config to use `javascript.formatter.trailingCommas`


Sec-ant (approved):
Thanks!

PR #651, opened 2024-06-22
feat: prefetch WASM file

Sec-ant:
## Summary

Prefetch the WASM file when user opens any page of the website, this can reduce the loading time of the playground.

### DIFF
--- a/src/components/starlight/Head.astro
+++ b/src/components/starlight/Head.astro
@@ -46,6 +46,6 @@ const isBlog = Astro.url.pathname.includes("/blog/");
 <script>
   import { prefetch } from "astro:prefetch";
   import WASM_URL from "../../../node_modules/@biomejs/wasm-web/biome_wasm_bg.wasm?url";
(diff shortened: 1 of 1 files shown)


ematipico:
Is there a particular reason why we should prefetch such a big file if not needed? I really don't mind the wait time, it's a playground after all 

Unfortunately we don't have metrics that could prove that the playground is our most visited page.

Sec-ant:
> Is there a particular reason why we should prefetch such a big file if not needed? I really don't mind the wait time, it's a playground after all
> 
> Unfortunately we don't have metrics that could prove that the playground is our most visited page.

I wouldn't say 3.2MB is such a big file. And it should be only fetched once and cached by the browser.

But yeah I think you're right, maybe it's only my personal preference to visit that page mostly. And the file size indicates that we probably don't need prefetching.

Anyway the code doesn't work as expected, the deploy preview shows the WASM file is still fetched when navigating to the playground. So I'll close this.

PR #662, opened 2024-06-24
docs: update lefthook example

AkaraChen:
## Summary

If the `root` of the lefthook command is not specified, `biome check` will not follow `biome.json[overrides]`

### DIFF
--- a/src/content/docs/ja/recipes/git-hooks.mdx
+++ b/src/content/docs/ja/recipes/git-hooks.mdx
@@ -23,6 +23,7 @@ Gitリポジトリのルートに`lefthook.yml`というファイルを追加し
   pre-commit:
     commands:
       check:
+        root: .
         glob: "*.{js,ts,cjs,mjs,d.cts,d.mts,jsx,tsx,json,jsonc}"
         run: npx @biomejs/biome check --no-errors-on-unmatched --files-ignore-unknown=true {staged_files}
   ```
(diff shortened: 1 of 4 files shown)


ematipico:
@nhedger is the change make sense to you? You used lefthook before

nhedger:
I've never used `root` before, and I'm not sure what the issue is in the first place. I'd like to reproduce the issue to confirm that this is indeed a documentation issue, and not a bug in Biome itself.

Would you be willing to provide a reproduction @AkaraChen ?

AkaraChen:
Here's the step by step guide to reproduce:

https://github.com/AkaraChen/biomejs-repro-662

@nhedger

nhedger:
Thanks for the detailed reproduction; this helped a lot!

I can reproduce the issue by following your instructions, but I suspect that the pattern specified in `includes` does not match the files in `dir-1`, which would explain why the file still gets linted.

Using the following value as `include` seems to correctly prevent Biome from listing the files in `dir-1`: `./dir-1/*`.

What I can't yet explain, though, is why setting the `root` setting in lefthook somehow makes it work.

---

@ematipico do we allow specifying folders as-is in the includes, or is it mandatory to specify a glob that matches _files_.

Sec-ant:
Here're some relative discussions on this glob match behavior: https://github.com/biomejs/biome/issues/2131#issuecomment-2011873329

ematipico:
> What I can't yet explain, though, is why setting the root setting in lefthook somehow makes it work.

It's possible that with this setting, lefthook computes the paths differently, which results in matching the glob specified. 

However, if there's an issue, it's definitely around globs and not how lefthook should be set up.

PR #674, opened 2024-06-25
Include links in badges

jamiebuilds:
## Summary

If people are wanting to include these badges in their readmes it's because they think it's important that people are familiar with the toolchain that's being used, in case people are not familiar a link is probably what they want

### DIFF
--- a/src/content/docs/recipes/badges.mdx
+++ b/src/content/docs/recipes/badges.mdx
@@ -7,7 +7,7 @@ Add the Biome badges to your `README.md`! 💅
 
 ## Formatter badge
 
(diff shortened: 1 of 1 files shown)


jamiebuilds:
I'll also suggest that people might want a more generic title if they want to adopt everything Biome has to offer:

- [Powered by Biome](https://biomejs.dev/)
- [Built with Biome](https://biomejs.dev/)

Sec-ant:
This page is a `.mdx` file, maybe we can add a component to let users customize the badge, with a few common presets.

Sec-ant (approved):

ematipico (approved):

PR #719, opened 2024-07-03
docs: modify lefthook re-stage example

u1aryz:
## Summary

lefthook has an option (`stage_fixed`) for re-staging, and I think this setting is more concise and intuitive.
https://github.com/evilmartians/lefthook/blob/master/docs/configuration.md#stage_fixed


Sec-ant (approved):
Thank you

PR #722, opened 2024-07-05
Fix overrides syntax in useFilenamingConvention example

venables:
Adds the missing `rules` key

## Summary

The overrides syntax has a `rules` key that was missing in this example.


ematipico:
Thank you @venables , but this document is auto-generated from source code

PR #770, opened 2024-07-14
docs: update auto sort on save vscode extension settings

KuroXI:
## Summary

This change is to clarify and correct the configuration settings related to auto-sorting imports on save within VS Code. The existing configuration had an incorrect key `source.organizeImports.biome`, which has been updated to the correct key `source.sortImports`.


ematipico:
Why is it incorrect? That's the exact key code that the Biome LSP needs to apply the import sorting

KuroXI:
You're correct. It seems like it doesn't organize `import for side effects`. Also, the `source.sortImports` action is using VSCode's built-in capabilities, even though the `defaultFormatter` is set to Biome.

PR #811, opened 2024-07-18
refactor: introduce the configure-biome guide

Conaclos:
## Summary

This PR renames `How Biome works` into `Configure Biome`.
I tried to improve the introduction of the Configuration file by talking about its structure.
We also point to the configuration reference for users looking for that.

I introduced a new section `Ignore files` that is the old `include and ignore explained`, but hopefully improved.
I first wanted separate this section into a dedicated guide.
However, I am no longer convinced it is a valuable change.
Also, I turn the `Protected files` section into a note in  `Ignore files`.

I merged the two sections about the `extends` field into `Share a configuration file`.

I removed duplicated docs in the formatter/linter intro by pointing to the `Ignore files` section of the new guide.

In the future, I think we could introduce a new section for explaining our `overrides` field.


nhedger:
I'm wondering if the title shouldn't be "Configur**ing** Biome" instead to be aligned with other section titles such as "Gett**ing** started".

ematipico:
Looking at Astro: https://docs.astro.build/en/install-and-setup/

They use "Getting started", but the rest of the guides don't use the "ing" form. For example, it is "Migrate to Astro" and not "Migrating to Astro"

Conaclos:
I have no particular preferences. I was also hesitating between `Ignoring files` and `Ignore file`. Maybe we should always use the infinitive form, but for `Getting Started` because this carries an action?

anonrig (approved):

nhedger:
No strong feelings either. It just came to mind

src/content/docs/guides/configure-biome.mdx
```diff
@@ -0,0 +1,391 @@
...
+This ensures that each team member has the same configuration in the CLI and in any editor that allows Biome integration.
+Many of the options available in a configuration file are also available in the CLI.
+
+## Configuration file structure
+
+A Biome configuration file is named `biome.json` or `biome.jsonc`.
+It is usually placed in your project's root directory, next to your project's `package.json`.
+
+Because Biome is a toolchain, its configuration is organized around the tools it provides.
+At the moment, Biome provides three tools: the formatter, the linter and the import organizer (also called the import sorter).
```
ematipico, line 23: It's the other way around. We take the name from the LSP, which calls it `organizeImports`. So for us is called import sorter first and foremost.

src/content/docs/guides/configure-biome.mdx
```diff
@@ -0,0 +1,391 @@
...
+  "linter": {
+    "enabled": false
+  },
+  "organizeImports": {
+    "enabled": false
+  }
+}
+```
+
+Options that apply to any language are placed in the corresponding tool field.
```
ematipico, line 42:
```suggestion
Options that apply to more than one language are placed in the corresponding tool field.
```

I suggest this because we have https://biomejs.dev/reference/configuration/#formatterattributeposition, which applies to JSX and HTML

src/content/docs/guides/configure-biome.mdx
```diff
@@ -0,0 +1,391 @@
...
+    "formatter": {
+      "enabled": false
+    }
+  }
+}
+```
+
+:::note
+Biome refers as `javascript` all variants of the JavaScript language.
+This includes TypeScript and JSX.
```
ematipico, line 72:
```suggestion
This includes TypeScrip, JSX and TSX.
```

I learned that some people think JSX and TSX are two separate languages.

src/content/docs/guides/configure-biome.mdx
```diff
@@ -0,0 +1,391 @@
...
+- Biome commands that run in `app/backend/package.json` will use the configuration file `app/backend/biome.json`;
+- Biome commands that run in `app/frontend/legacy/package.json` and `app/frontend/new/package.json`
+will use the configuration file `app/frontend/biome.json`;
+
+:::caution
+Biome doesn't support nested `biome.json` files, neither in CLI nor in LSP. [Follow and help the support in the related issue](https://github.com/biomejs/biome/issues/2228)
+:::
+
+:::note
+Biome commands support using the `--config-path` option.
```
ematipico, line 108: I think we should mention the env variable `BIOME_CONFIG_PATH` too

src/content/docs/guides/configure-biome.mdx
```diff
@@ -0,0 +1,391 @@
...
+`include` and `ignore` accepts globs patterns such as `src/**/*.js`.
+See the related section for which glob syntaxes are supported.
+`include` is always applied first before applying `ignore`.
+This allows you to include some files and to ignore some of the file you included.
+
+:::note
+`include` and `ignore` have a slightly different semantics.
+`ignore` applies to both regular files and folders,
+while `include` is only tested on regular files.
+This means that if you want to prevent Biome from traversing a folder,
```
ematipico, line 200: Are you sure this is true? I used `include` for folders, and it worked

src/content/docs/guides/configure-biome.mdx
```diff
@@ -0,0 +1,391 @@
...
+Some examples:
+
+- `dist/**` matches the dist directory and all files in this directory.
+- `**/test/**` matches all files under any directory named `test`, regardless of where they are. E.g. `dist/test`, `src/test`.
+- `**/*.js` matches all files ending with the extension `.js` in all directories.
+
+Biome uses a glob library that treats all globs as having a `**/` prefix.
+This means that `src/**/*.js` and `**/src/**/*.js` are treated as identical.
+They match both `src/file.js` and `test/src/file.js`.
+This is something we plan to fix in Biome **v2.0.0**, as it will be a breaking change.
```
ematipico, line 384: There's no need to mention "future" plans inside docs:
- it forces us to remember to remove this line when there's a new release
- it "expires" and confuses people
- feels like a `TODO`

When the logic changes, we change the explanation accordingly.

src/content/docs/guides/configure-biome.mdx
```diff
@@ -0,0 +1,391 @@
...
+`include` is always applied first before applying `ignore`.
+This allows you to include some files and to ignore some of the file you included.
+
+:::note
+`include` and `ignore` have a slightly different semantics.
+`ignore` applies to both regular files and folders,
+while `include` is only tested on regular files.
+This means that if you want to prevent Biome from traversing a folder,
+you have to add the folder to `ignore`.
+This is a current limitation that we will fix in a future release.
```
ematipico, line 202: To remove. No need to explain future stuff, because we don't know when it will change and how. And it forces us to come back here and remove it.

ematipico (approved):
Awesome work! I added few nits, but overall I like the new change! ❤️

src/content/docs/guides/configure-biome.mdx
Conaclos: I copied this from the old docs. Maybe it is badly phrased. It means to say that `include` doesn't prevent Biome of traversing a folder that is not covered by `include`.
For example, if `include` contains only `src`, Biome will still visit `test`.

Conaclos: This should be fixed :)

arendjr:
Regarding title verbs, I would use the -ing forms. It appears to be the most common choice, but style guided may differ of course. “Getting Started” is the most common example, and it feels less like a command than the imperative form “Get Started”. To me, “Configure Biome” feels like an action that should be taken, like a call-to-action button. But the content of the chapter is a guide on “configuring Biome”.

Found a good discussion on the topic also:

https://writing.stackexchange.com/questions/1545/verb-tense-for-technical-document-titles

I think the strongest argument I could find for using the imperative form (as is done currently in this PR) is that these forms of the verbs might be easier to recognize for non-English speakers. I think the strongest argument for continuous form (apart from historical convention) is that it is easiest to do consistently, since people note various examples where imperative form just feels awkward, so they still end up using “Getting Started” or the example of “Configuring Kubernetes”, despite using imperative form elsewhere).

Anyway, just food for thought :)

arendjr (approved):

Conaclos:
I am going to merge the PR. Please feel free to open new PRs to update translations.

Issue #872, opened 2024-07-31
📚 I want to add `--colors=off` option to Lefthook example


torabit:
### Documentation URL

https://biomejs.dev/ja/recipes/git-hooks/

### Description

The git error log output by Lefthook outputs ANSI escape code as is, so we need to add an option to display plain text.

```mdx
> git -c user.useConfigOnly=true commit --quiet --allow-empty-message --file -
╭───────────────────────────────────────╮
│ 🥊 lefthook v1.7.11  hook: pre-commit │
╰───────────────────────────────────────╯
sync hooks: ✔️ (pre-commit)
┃  check ❯ 

[0m.//src/lib/Foo/Foo.tsx[0m[0m:[0m[0m4[0m[0m:[0m[0m34[0m[0m [0m[0mparse[0m[0m [0m[0m━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━[0m[0m
...
```
### Expectations

It took me a little while to get to this information, but wouldn't it be more helpful to include it by default in the Lefthook Example in the Git Hook section?

```mdx
```yaml title="lefthook.yml"
pre-commit:
  commands:
    check:
      glob: "*.{js,ts,cjs,mjs,d.cts,d.mts,jsx,tsx,json,jsonc}"
      run: npx @biomejs/biome check --no-errors-on-unmatched --files-ignore-unknown=true --colors=off {staged_files}
```

P.S
I was going to make a PR but I didn't have the authority to do so 😿.

### Code of Conduct

- [X] I agree to follow Biome's Code of Conduct

Conaclos:
> I was going to make a PR but I didn't have the authority to do so 😿.

What do you mean? Anyone is able to do a PR to this repository?

torabit:
@Conaclos 
Only those in authority can make Pull Request.

ematipico:
> @Conaclos 
> > What do you mean? Anyone is able to do a PR to this repository?
> 
> No, Only those in authority can make Pull Request.

You have to **fork** the repository, and you can create a new branch, apply the new changes, and send a PR to this repository.

torabit:
@ematipico 

Oh! I was able to do it!

This was the first time in my life that I had ever been a contributor, so I didn't know the etiquette 🙈 .

Thank you very much.

ematipico:
Fixed by https://github.com/biomejs/website/pull/876

[...]

PR #876, opened 2024-08-01
docs(Git Hooks): add colors option for lefthook example :memo:

torabit:
## Summary
https://github.com/biomejs/website/issues/872

> The git error log output by Lefthook outputs ANSI escape code as is, so we need to add an option to display plain text.
> It took me a little while to get to this information, but wouldn't it be more helpful to include it by default in the Lefthook Example in the Git Hook section?


Conaclos:
Thanks! Congratulations on your first contribution to Open Source projects :)

PR #975, opened 2024-08-22
feat: make heading sizes smaller

nhedger:
## Summary

This is an experiment to see whether reducing the size of the headings improves readability.

### DIFF
--- a/src/styles/_patches.scss
+++ b/src/styles/_patches.scss
@@ -5,18 +5,31 @@
 	line-height: var(--sl-line-height-headings);
 }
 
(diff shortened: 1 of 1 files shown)


ematipico:
I believe the change is fine, however this also changed the font of "On this page", which you usually find on the right side bar. I think it's fine to keep that font size as it was

nhedger:
Ah, I hadn't realized. It makes sense to keep its original size, indeed.

ematipico:
I believe we can merge the PR

PR #1010, opened 2024-08-28
feat: add support for html, vue, and svelte files in the playground

### LINKED ISSUE #1009: Playground doesn't allow you to name a file with an extension like `.svelte`
### Description

When you try to create or rename a file in the playground (multi file mode) with an extension like `.svelte`, the playground normalizes it to `.svelete.js`. This doesn't allow us to test features in such files.

### Expectations

The extension should not be normalized. It should be left intact.

### Code of Conduct

dyc3:
## Summary

fixes #1009

### DIFF
--- a/src/playground/workers/prettierWorker.ts
+++ b/src/playground/workers/prettierWorker.ts
@@ -14,16 +14,23 @@ import {
 import {
 	isCssFilename,
 	isGraphqlFilename,
+	isHtmlFilename,
 	isJsonFilename,
+	isSvelteFilename,
 	isTypeScriptFilename,
+	isVueFilename,
 } from "@/playground/utils";
 import * as prettier from "prettier";
+// @ts-expect-error
+import * as pluginSvelte from "prettier-plugin-svelte/browser";
 // @ts-expect-error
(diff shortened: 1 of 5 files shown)


dyc3:
The Svelte prettier plugin issue seems to be an upstream problem: prettier-plugin-svelte#69. So we aren't doing anything wrong here.

src/playground/workers/prettierWorker.ts
```diff
@@ -109,7 +116,14 @@ async function formatWithPrettier(
 			tabWidth: options.indentWidth,
 			printWidth: options.lineWidth,
 			filepath: options.filepath,
-			plugins: [parserBabel, pluginEstree, pluginCss, pluginGraphql],
+			plugins: [
+				parserBabel,
+				pluginEstree,
+				pluginCss,
```
chansuke, line 122: nit: sorting alphabetically is better for readability maybe?
```suggestion
				pluginCss,
				pluginEstree,
```

chansuke (approved):
Overall, good for me. Just added a comment.

PR #1012, opened 2024-08-28
fix(wasm): regression

ematipico:
## Summary

It's weird that we didn't catch this.

Anyway, this should fix the issue


src/playground/workers/biomeWorker.ts
```diff
@@ -40,6 +40,8 @@ let configuration: undefined | Configuration;
 function getPathForFile(file: File): BiomePath {
 	return {
 		path: file.filename,
+		kind: ["Handleable"],
```
dyc3, line 43: As a quick fix, this is functional
```suggestion
		kind: 1 << 4,
```

dyc3 (approved):

PR #1089, opened 2024-09-11
fix: handle unimplemented file capabilities slightly more gracefully

dyc3:
## Summary

The HTML language currently doesn't implement some of the file capabilities that the playground expects. This makes it so that a language does not need to implement some of these capabilities in order to work in the playground, requiring less code changes. Ideally though, we would grab the file capabilities from biome instead of try/catching.

This doesn't make HTML fully functional in the playground -- other things in biome needs to be implemented first.

### DIFF
--- a/src/playground/workers/biomeWorker.ts
+++ b/src/playground/workers/biomeWorker.ts
@@ -14,6 +14,7 @@ import {
 	isCssFilename,
 	isFrameworkTemplateFilename,
 	isGraphqlFilename,
+	isHtmlFilename,
 	isJsonFilename,
 } from "@/playground/utils";
 import init, {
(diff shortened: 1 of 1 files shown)


ematipico:
> Ideally though, we would grab the file capabilities from biome instead of try/catching.

This is actually possible. The workspace provides a method for that

dyc3:
Ah, then I will rework this to do that.

dyc3:
It looks like the `fileFeatures` function doesn't currently indicate if the language supports parsing, control flow graph, or formatter ir.

dyc3:
Now that the other PRs in biome are merged, everything should be good to go here.

src/playground/workers/biomeWorker.ts
```diff
@@ -278,9 +279,12 @@ self.addEventListener("message", async (e) => {
...
 
-			const printed = workspace.formatFile({
-				path,
-			});
+			const printed =
+				fileFeatures.features_supported.get("Format") === "Supported"
+					? workspace.formatFile({
+							path,
+						})
+					: { code };
```
Princesseuh, line 287: Should this also return a `Not supported` string?

ematipico (approved):
Looks good. I would use the message `"Not supported"` instead. The word "available" doesn't fit the case, I think. I would address  @Princesseuh 's comment

PR #1100, opened 2024-09-12
Update annoucing-biome.mdx

richardtorres314:
## Summary
Updating punctuation on page


dyc3 (commented):
The comma is correct punctuation. Semi colons are typically used in place of words like "and" or "but", separating 2 sentences.

PR #1186, opened 2024-09-21
docs(ja): translate third-party actions

toririm:
## Summary

Translates a section "Third-party actions" to Japanese.


unvalley (approved):
Thank you!

PR #1218, opened 2024-09-26
fix: `biome check` error

chansuke:
## Summary

Fixed `biome check` error on main branch.

- any type error
- formatting


ematipico:
Thank you for fixing my mistake

PR #1238, opened 2024-10-02
fix: codegen of some rule pages

ematipico:
## Summary

The code gen was broken for some edge case coming from our markdown. 

This PR fixes it.

I confirmed locally that the website builds


codegen/src/lintdoc.rs
```diff
@@ -591,7 +591,11 @@ fn write_documentation(
                     write!(block, "{text}")?;
                 }
 
-                write!(content, "{text}")?;
+                if matches!(text.as_ref(), "`" | "*" | "_") {
+                    write!(content, "\\{text}")?;
+                } else {
+                    write!(content, "{text}")?;
+                }
```
ematipico, line 598: This is the fix

ematipico:
It built, awesome!

PR #1244, opened 2024-10-03
Add Deno installation script

mezotv:
## Summary
Adds the deno install script according to its [documentation](https://docs.deno.com/runtime/reference/cli/install/) to have support for all the package managers.

It does not appear that deno has an equivalent to the --exact flag.


ematipico (changes requested):
The change [doesn't work](https://deploy-preview-1244--biomejs.netlify.app/guides/getting-started/#installation), probably because the component doesn't accept a `deno` prop. 

Also, would you mind updating the other `getting-started.mdx` pages?

mezotv:
> The change [doesn't work](https://deploy-preview-1244--biomejs.netlify.app/guides/getting-started/#installation), probably because the component doesn't accept a `deno` prop.
> 
> Also, would you mind updating the other `getting-started.mdx` pages?

Will do give me a second!

mezotv:
There we go!

image

ematipico (approved):

PR #1246, opened 2024-10-04
fix: example config was using a deprecated key

mountainash:
## Summary

`The property trailingComma is deprecated. Use javascript.formatter.trailingCommas instead. biome(deserialize)`

This commit updates the code sample to be up-to-date.

Screenshot showing the warning in VScode


ematipico (approved):
Thank you

PR #1316, opened 2024-10-17
Better instructions for import sorting

jamesliu4c:
- Instructions would have done all formatting and safe lining fixes before.
- Use `--write` instead of `--apply` because `--apply` has been deprecated in favor of `--write`.

## Summary

Instructions would have done all formatting and safe lining fixes before this change. Now, it describes how to just apply the formatting changes.

Use `--write` instead of `--apply` because `--apply` has been deprecated in favor of `--write`.

### DIFF
--- a/src/content/docs/analyzer/import-sorting.mdx
+++ b/src/content/docs/analyzer/import-sorting.mdx
@@ -147,10 +147,15 @@ import React from "react";         // Import group 4
 
 ## Import sorting via CLI
 
-Using the command `check`, with the option `--apply`.
+Using the command `check`, with the option `--apply`. If you want only order the imports, you can use check like so:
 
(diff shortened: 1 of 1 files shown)


src/content/docs/analyzer/import-sorting.mdx
```diff
@@ -147,10 +147,15 @@ import React from "react";         // Import group 4
 
 ## Import sorting via CLI
 
-Using the command `check`, with the option `--apply`.
+Using the command `check`, with the option `--apply`. If you want only order the imports, you can use check like so:
```
ematipico, line 150:
```suggestion
Using the command `check`, with the option `--write`. If you want only order the imports, you can use check like so:
```

PR #1327, opened 2024-10-20
style: simplify string formatting for readability

### LINKED ISSUE #1326: `writeln!` and `format!` code can be simplified
There are about a dozen or so instances where we can do something like

```rs
writeln!(content, "{HEADER}")?;
```

instead of

```rs
    writeln!(content, "{}", HEADER)?;
```

to improve the codebase's readability and maintainability.

hamirmahal:
closes #1326.


hamirmahal (commented):
`cargo test` passes locally.

PR #1430, opened 2024-11-12
feat: narrow types down

thilllon:
## Summary

The type of tab key is currently a `string`, and `formatter` is being used as a magic word. To make it easier for contributors to understand, I've made the type more strict and replaced the magic word with an enum.


src/playground/types.ts
```diff
@@ -2,6 +2,18 @@ import type { Diagnostic } from "@biomejs/wasm-web";
 import type { parser } from "codemirror-lang-rome-ast";
 import type { Dispatch, SetStateAction } from "react";
 
+export enum TabKeys {
```
dyc3, line 5: I would prefer if this wasn't an enum. I think it would be better if this was a union type instead.

dyc3, line 5: nit: perhaps `PlaygroundTab` would be a better type name

dyc3 (changes requested):

dyc3 (approved):

PR #1538, opened 2024-12-14
fix(index): make progress colors more accessible

illright:
## Summary

I found that the existing colors of progress bars on the index page were very bright and didn't have sufficient contrast with the text on them, which made that text hard to read. I propose a different set of colors and also increased font weight for durations to reach contrast targets set by WCAG.

Before (https://webaim.org/resources/contrastchecker/?fcolor=353841&bcolor=FFA07A):
image

After (https://webaim.org/resources/contrastchecker/?fcolor=353841&bcolor=FFA07A):
image

-----
<a href="https://stackblitz.com/~/github.com/illright/website/tree/illright/patch-52892">Review PR in StackBlitz</a> _Submitted with [StackBlitz](https://stackblitz.com/~/github.com/illright/website/tree/illright/patch-52892)._

### DIFF
--- a/src/content/docs/index.mdx
+++ b/src/content/docs/index.mdx
@@ -73,8 +73,8 @@ import AwardBanner from "@/components/AwardBanner.astro"
                         <ProgressBarContainer
                             client:visible
                             data={[
-                                {duration: 0.41, label: "Biome", color: "#33ff66"},
-                                {duration: 14.35, label: "Prettier", color: "#ff0033"}
+                                {duration: 0.41, label: "Biome", color: "MediumSeaGreen"},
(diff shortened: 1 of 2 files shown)


dyc3 (approved):

dyc3 (changes requested):
Ah, actually this regresses the dark mode version.

image

illright:
@dyc3 ah, good catch. Slightly revised the solution, see updated PR description

ematipico (commented):
We have an index for each language, we would appreciate it if you could update them too

ematipico (approved):

dyc3 (approved):

PR #1560, opened 2024-12-19
docs: update version notice in noRestrictedImports

MaxtuneLee:
## Summary

Changed the version number from `v1.6.0` to `v2.0.0` to reflect the correct introduction version of the `no-restricted-imports` rule.

Issue mentioned in https://github.com/biomejs/biome/issues/4750


ematipico:
@MaxtuneLee this is auto-generated file, as mentioned at the top of the file.

You'll have to change the source code in `biomejs/biome`

PR #1615, opened 2024-12-31
docs(ja): translate gritql page

mehm8128:
## Summary

- translate gritql page to ja
- fixed the command on README for the change of https://github.com/biomejs/website/pull/1540


src/content/docs/ja/reference/gritql.mdx
```diff
@@ -0,0 +1,108 @@
...
+
+GritQL クエリは _パターン_ を通じて機能します。
+最もよく見られるパターンは、バッククォートで囲まれた通常のソースコードのように見えるコードスニペットです。
+
+```grit
+`console.log('Hello, world!')`
+```
+
+このパターンは、文字列 `'Hello, world!'` が渡された全ての `console.log()` の呼び出しに一致します。
+ただし、GridQLは _構造的_ マッチングを行うため、フォーマットの詳細には関心がありません。
```
mehm8128, line 27: textlint says we prefer to use「format」 to 「フォーマット」, but I think it is not proper for this sentence. Is there any idea?

unvalley: yeah, I think we should use リント and フォーマット instead of alphabet in overall now. (Initially I thought that リント is a bit awkward, so I decided to use lint and format as alphabet because they are also technical words.)

Could you fix it to pass textlint ci for now? I will adjust everything later.

unvalley (approved):
LGMT if resolved the comment

src/content/docs/ja/reference/gritql.mdx
mehm8128: I understand, thank you!
I fixed temporarily it to be "format"!

unvalley:
Thank you!

PR #1660, opened 2025-01-06
fix(docs): Put noConsole rule in the correct category

ofhouse:
## Summary

According to the [latest schema](https://biomejs.dev/schemas/1.9.4/schema.json) the `noConsole` log rule is located under `nursery` category and not under `suspicious`.

**Changes**
- Replaces `suspicious` category with `nursery`


ematipico:
The schema is incorrect.

The rule is under the correct group: https://github.com/biomejs/biome/blob/main/crates%2Fbiome_js_analyze%2Fsrc%2Flint%2Fsuspicious%2Fno_console.rs

Also, the documentation is generated from the source (from the file I pasted)

nhedger:
Looks like the schema is correct now.

I'll close this, but thanks for your time anyway, @ofhouse!

PR #1663, opened 2025-01-08
docs: add French translation of reference/cli.mdx

webdevbynight:
## Summary

Add the translation into French of the CLI page.

Added:
- the `reference/cli.mdx` file translated into French.


ematipico:
This file is auto generated, so we can't translate it :(

PR #1665, opened 2025-01-08
Clarify conditions under which use-exhaustive-dependencies.mdx can be used

tommy-gilligan:
## Summary

`only` is a strong word.  If later we refer to Preact in addition to React, this should be included when stating conditions under which this rule can be used.


dyc3:
As stated at the top of this file, this page is auto generated from the source code of the rule in the main biome repo. Make the changes in there instead.

PR #1666, opened 2025-01-09
Commercial support page

arendjr:
## Summary

This adds a page for commercial support to our website:

commercial_support

The page is also linked to from the front page (I've removed the rather non-specific "Optimized" section):

commercial_support_card

Note that the link in the card for this doesn't work yet. I suspect it's a CSS issue, but I didn't yet see the source of the issue :thinking: 

Let me know what you think!

### DIFF
--- a/astro.config.ts
+++ b/astro.config.ts
@@ -94,6 +94,11 @@ export default defineConfig({
 					},
 				},
 				{
+					label: "Support",
+					link: "../support",
+					translations: {},
+				},
+				{
 					label: "Guides",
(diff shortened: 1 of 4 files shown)


ematipico:
Screen Shot 2025-01-09 at 12 22 43


On mobile, I would stack the layout:
- the two buttons should go one below the other
- we could hide the heart icon, or put it *below* the test

arendjr:
Responsiveness is fixed :+1:

ematipico (approved):

PR #1674, opened 2025-01-13
feat: add "How to configure" in rules

ematipico:
## Summary

This PR adds:
- A new header `## Description` before the description of the rule
- a new `## How to configure` header inside the page of a single rule, after the examples and before the related links. The paragraph contains how a user can configure a rule. I added this because a (new) user didn't know how to opt-in a non-recommended rule from the page. This should help


codegen/src/lintdoc.rs
```diff
@@ -8,7 +8,7 @@ use biome_analyze::{
     AnalysisFilter, AnalyzerOptions, ControlFlow, FixKind, GroupCategory, Queryable,
     RegistryVisitor, Rule, RuleCategory, RuleFilter, RuleGroup, RuleMetadata, RuleSourceKind,
 };
-use biome_configuration::PartialConfiguration;
+use biome_configuration::{ PartialConfiguration};
```
arendjr, line 11: :suspect:

arendjr (approved):

PR #1677, opened 2025-01-14
Fix links in cards

arendjr:
## Summary

I forgot to fix the "Commercial support" link in the card on the homepage. It works now.


ematipico (approved):

ematipico:
This feels like a bug to me. Should we report it to starlight?

PR #1701, opened 2025-01-22
feat(playground): add tab to preview analyzer fixes

### LINKED ISSUE #1052: Add options to apply safe and unsafe lint fixes to the playground
The playground is a great way for users to share formatting and lint rule bugs with us. However, we currently don't have any way to preview automatic lint fixes in the playground. It would be nice to have a couple of options available to apply safe and unsafe fixes.

Issues like this would benefit from having this available:
- biomejs/biome#3790

dyc3:
## Summary

This PR adds a new tab to preview analyzer fixes. Open to suggestions for naming things.

closes #1052

### DIFF
--- a/src/playground/Playground.tsx
+++ b/src/playground/Playground.tsx
@@ -48,6 +48,7 @@ import {
 	useRef,
 	useState,
 } from "react";
+import AnalyzerFixesTab from "./tabs/AnalyzerFixesTab";
 
 export default function Playground({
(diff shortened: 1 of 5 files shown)


ematipico:
It seems that only safe fixes are applied. Should we provide an option for applying unsafe fixes too?

dyc3:
Yup, that's been added to this PR, along with "Apply suppressions"

PR #1731, opened 2025-02-10
docs: add GitLab CI configuration example to continuous integration g…

coka-stefan:
## Summary

Added example for usage in GitLab CI pipelines

### DIFF
--- a/src/content/docs/recipes/continuous-integration.mdx
+++ b/src/content/docs/recipes/continuous-integration.mdx
@@ -71,7 +71,8 @@ lint:
       entrypoint: [""]                    # This is required for the image to work in GitLab CI
     stage: quality                        # Run in the quality stage
     script:
-        - biome ci src --reporter=gitlab --colors=off --log-level=none --diagnostic-level=error > code-quality.json  # Execute Biome on the 'src' directory
+        - biome ci --reporter=gitlab --colors=off > /tmp/code-quality.json 
+        - cp /tmp/code-quality.json code-quality.json
     artifacts:
         paths:
             - code-quality.json    # Collect the code quality report as an artifact
--- a/src/content/docs/recipes/continuous-integration.mdx
+++ b/src/content/docs/recipes/continuous-integration.mdx
@@ -82,4 +82,6 @@ lint:
     allow_failure: true            # Allow the job to fail (adjust as needed)
 ```
 
-> **Note:** If your project's source code is located in another directory, replace `src` with that directory's path in the command.
+:::note
+If your project's source code is located in another directory, replace `src` with that directory's path in the command.
+:::
--- a/src/content/docs/recipes/continuous-integration.mdx
+++ b/src/content/docs/recipes/continuous-integration.mdx
@@ -66,17 +66,20 @@ stages:
 
 # Lint job configuration: runs code quality checks using Biome
 lint:
-    image: node:lts         # Use the latest Node.js LTS image
-    stage: quality          # Run in the quality stage
-    before_script:
-        - npm install       # Install dependencies using npm
+    image: 
+      name: ghcr.io/biomejs/biome:latest  # Use the latest Biome Docker image
+      entrypoint: [""]                    # This is required for the image to work in GitLab CI
+    stage: quality                        # Run in the quality stage
     script:
-        - npm run biome ci src --reporter=summary  # Execute Biome using npx on the 'src' directory
-    needs: []               # No dependent jobs needed
-    cache:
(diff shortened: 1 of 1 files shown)


nhedger:
We actually ship Docker images for Biome which can be used in GitLab CI. Maybe the example could use them instead. Save the user from installing Biome in a node image.

src/content/docs/recipes/continuous-integration.mdx
```diff
@@ -54,3 +54,29 @@ jobs:
...
+  - quality
+
+# Lint job configuration: runs code quality checks using Biome
+lint:
+    image: node:lts         # Use the latest Node.js LTS image
+    stage: quality          # Run in the quality stage
+    before_script:
+        - npm install       # Install dependencies using npm
+    script:
+        - npm run biome ci src --reporter=summary  # Execute Biome using npx on the 'src' directory
```
ematipico, line 74: Biome provides a gitlab reporter, shouldn't we use that instead?

coka-stefan: Ah, I was not aware of this, as we started using biome in gitlab before the gitlab reporter was implemented.

src/content/docs/recipes/continuous-integration.mdx
```diff
@@ -54,3 +54,32 @@ jobs:
...
+        paths:
+            - code-quality.json    # Collect the code quality report as an artifact
+    needs: []                      # No dependent jobs needed
+    rules:
+        - if: $CI_COMMIT_BRANCH    # Run job for commits on branches
+        - if: $CI_MERGE_REQUEST_ID # Run job for merge requests
+    allow_failure: true            # Allow the job to fail (adjust as needed)
+```
+
+> **Note:** If your project's source code is located in another directory, replace `src` with that directory's path in the command.
```
ematipico, line 85:
```suggestion
:::note
If your project's source code is located in another directory, replace `src` with that directory's path in the command.
:::
```

ematipico (approved):
Looks good to me. @nhedger please check the docker side and merge it whenever you see fit :)

src/content/docs/recipes/continuous-integration.mdx
```diff
@@ -54,3 +54,34 @@ jobs:
...
+  - quality
+
+# Lint job configuration: runs code quality checks using Biome
+lint:
+    image: 
+      name: ghcr.io/biomejs/biome:latest  # Use the latest Biome Docker image
+      entrypoint: [""]                    # This is required for the image to work in GitLab CI
+    stage: quality                        # Run in the quality stage
+    script:
+        - biome ci src --reporter=gitlab --colors=off --log-level=none --diagnostic-level=error > code-quality.json  # Execute Biome on the 'src' directory
```
nhedger, line 74:
```suggestion
        - biome ci --reporter=gitlab > code-quality.json
```

Perhaps the example should be less opinionated unless these additional params have been added so that exporting code-quality.json works as expected. You probably have tested this better than I have, so I'll trust your jugement on this.

Regardless, we probably don't want to refer to `src`, because there may be supported files somewhere else in the project

src/content/docs/recipes/continuous-integration.mdx
```diff
@@ -54,3 +54,34 @@ jobs:
...
+    artifacts:
+        paths:
+            - code-quality.json    # Collect the code quality report as an artifact
+    needs: []                      # No dependent jobs needed
```
nhedger, line 78:
```suggestion

```

If not required perhaps we could do without it in the example

src/content/docs/recipes/continuous-integration.mdx
```diff
@@ -54,3 +54,34 @@ jobs:
...
+    rules:
+        - if: $CI_COMMIT_BRANCH    # Run job for commits on branches
+        - if: $CI_MERGE_REQUEST_ID # Run job for merge requests
+    allow_failure: true            # Allow the job to fail (adjust as needed)
```
nhedger, line 82:
```suggestion

```

Perhaps we remove this one too. I expect people who use GitLab CI on a daily basis to know about this option. It keeps the example leaner. What do you think ?

src/content/docs/recipes/continuous-integration.mdx
coka-stefan: When running `biome ci --reporter=gitlab > code-quality.json` the exit code is 1 because biome runs CI on the `code-quality.json` as well for some reason. I managed to avoid this issue, but the script is very hacky. Not sure if this should be used in the docs.

```yaml
script:
  - biome ci --reporter=gitlab --colors=off > /tmp/code-quality.json 
  - cp /tmp/code-quality.json code-quality.json
```

P.s. `--colors=off` is required because otherwise, the piped output is not a valid JSON.

ematipico (approved):

PR #1727, opened 2025-02-05
feat(playground): remove import sorting tab, rename import sorting to assist, add rule domains

dyc3:
## Summary

This fully removes the import sorting tab and makes the UI changes necessary to show "Assist" instead of "Import sorting".

It also adds UI for being able to specify rule domains. I'm not entirely sure what to do with the `Lint Rules` option though. Open to suggestions.

> [!CAUTION]
> This PR is stacked on the following PRs. **Do not merge it without merging the others and rebasing this branch.** To review this PR, look at the last commit only.
> - #1722

PR #1741, opened 2025-02-14
feat(playground): add html formatter options

dyc3:
## Summary

Adds HTML specific formatter options to the playground's settings. I've verified that they get passed into both biome and prettier's configuration. It won't be obvious it's working for biome's formatter because the logic for these options isn't entirely complete in biome.

> [!CAUTION]
> This PR is stacked on the following PRs. **Do not merge it without merging the others and rebasing this branch.** To review this PR, look at the last commit only.
> - #1727

### DIFF
--- a/src/playground/PlaygroundLoader.tsx
+++ b/src/playground/PlaygroundLoader.tsx
@@ -12,6 +12,7 @@ import {
 	type QuoteStyle,
 	type Semicolons,
 	type TrailingCommas,
+	type WhitespaceSensitivity,
 	defaultPlaygroundState,
 	emptyBiomeOutput,
 	emptyPrettierOutput,
@@ -247,6 +248,14 @@ function buildLocation(state: PlaygroundState): string {
 		}
 	}
 
+	// handle rule domains
+	for (const key in state.settings.ruleDomains) {
+		const value = state.settings.ruleDomains[key];
+		if (value !== undefined && value !== "none") {
+			queryStringObj[`ruleDomains.${key}`] = value;
+		}
+	}
+
 	const queryString = new URLSearchParams(queryStringObj).toString();
 	lastSearchStore.set(queryString);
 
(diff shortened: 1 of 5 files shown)


src/playground/PlaygroundLoader.tsx
```diff
@@ -301,6 +309,18 @@ function initState(
 		files = defaultPlaygroundState.files;
 	}
 
+	// handle rule domains
+	const ruleDomains: Record<string, string> = {};
+	for (const key of searchParams.keys()) {
+		if (key.startsWith("ruleDomains.")) {
+			const domain = key.slice("ruleDomains.".length);
+			const value = searchParams.get(key);
+			if (value !== null) {
```
chansuke, line 318: i believe you did this intentionally, but you might be able to simplify it as follows?:

```suggestion
			if (value) {
```

src/playground/PlaygroundLoader.tsx
```diff
@@ -301,6 +309,18 @@ function initState(
 		files = defaultPlaygroundState.files;
 	}
 
+	// handle rule domains
+	const ruleDomains: Record<string, string> = {};
+	for (const key of searchParams.keys()) {
+		if (key.startsWith("ruleDomains.")) {
+			const domain = key.slice("ruleDomains.".length);
```
chansuke, line 316: imho: you might consider moving `"ruleDomains.".length` outside the loop to avoid recalculating the slice length on every iteration.

```suggestion
        const prefixLength = "ruleDomains.".length;
	for (const key of searchParams.keys()) {
		if (key.startsWith("ruleDomains.")) {
			const domain = key.slice(prefixLength);
```

ematipico:
I believe we have to fix the scrolling issue we have on mobile (Firefox), because I can't see the new options 🫠

dyc3:
I've created #1743 to track the scrolling issue.

chansuke (approved):
lgtm

PR #1742, opened 2025-02-14
feat: revamp of the linter and assist documentation

ematipico:
## Summary

This is a big one, here's the highlight of the changes **user face**. The changes applied to the codegen merely reflect that changes applied to the structure of the new, proposed documentation.

Closes https://github.com/biomejs/website/issues/1765
Closes https://github.com/biomejs/website/issues/1764
Closes https://github.com/biomejs/website/issues/1766
Closes https://github.com/biomejs/website/issues/1767

### Rule page

I removed the "Note" and moved all the important information inside a `## Summary` header. The information is just a bullet list. I updated this list to also account for the new domains.

### Rules and actions divided by language

Rules and actions are divided by language. Now that our infrastructure is maturing and we are spanning rules across languages, the current division of the pages isn't optimal, and people might get confused when they see multiple instances on one page. So now, the page `/linter/rules/index.mdx` doesn't exist anymore, instead is copied under each language we support: `/linter/javascript/rules/index.mdx`, `/linter/css/rules/index.mdx` and so on.

You can see the result for yourselves, because it gets **way** easier to understand when a rule is applied to specific language. Sure, the icons did help, but I believe having a dedicated page reduces the cognitive load of our users. 

### Documentation of linter groups

I removed the code generation of the linter groups. It seemed unnecessary, considering they are hardcoded. Instead, I moved the groups' documentation inside the `/linter` page. Plus, I used lowercase names so they could reflect the names we have in the configuration. 

### Revamp of the linter page

I changed the documentation of the `/linter` page:
- Removed the `## Rules` header, and all the sub-headers are now a `##` as a result
- Documented the new suppression systems we have
- Added a FAQ section at the end
- Expanded the migration guide by using the new `--suppress` CLI option that we added in v2
- Removed all the `Since ..` from the web page. I rewrote everything by assuming that we have it in v2 since we don't maintain multiple versions of Biome.

### Expressive code goodies

I added a plugin to expressive code (it required a config), which shows the line numbers of the snippets. 

I needed those line numbers to document the new suppression system of the linter. I hope it's a good addition. It's still possible to disable the line numbers *per code block*.

## What's left

Linter domains haven't been documented. I will do that in a subsequent PR. I plan to have some codegen since there is "hidden" information that we have to pull, such as the dependencies of each domain.


src/content/docs/zh-CN/reference/configuration.mdx
```diff
@@ -266,7 +264,14 @@ Biome应检查VCS文件的文件夹。默认情况下，Biome将使用找到 `bi
...
 
-<LintGroups/>
+- accessibility: Rules focused on preventing accessibility problems.
+- complexity: Rules that focus on inspecting complex code that could be simplified.
+- correctness: Rules that detect code that is guaranteed to be incorrect or useless.
+- nursery: New rules that are still under development.  Nursery rules require explicit opt-in via configuration on stable versions because they may still have bugs or performance problems. They are enabled by default on nightly builds, but as they are unstable their diagnostic severity may be set to either error or warning, depending on whether we intend for the rule to be recommended or not when it eventually gets stabilized. Nursery rules get promoted to other groups once they become stable or may be removed.  Rules that belong to this group are not subject to semantic version.
+- performance: Rules catching ways your code could be written to run faster, or generally be more efficient.
+- security: Rules that detect potential security flaws.
+- style: Rules enforcing a consistent and idiomatic way of writing your code.
+- suspicious: Rules that detect code that is likely to be incorrect or useless.
```
chansuke, line 274: I might misunderstanding, do we plan to translate these bullet points into Chinese in the future?

ematipico: They weren't translated to begin with (you can check the current deployed website). I just replaced what we had with hardcoded text to not fail the build.

To answer your question, maybe? I'm not sure. Translations are hard without people committed to that, probably it's not worth that

chansuke: got it(I checked the current website), thanks for the explanation.

chansuke (approved):
lgtm

PR #1773, opened 2025-02-23
docs: Give explicit lspBin example

oliversalzburg:
## Summary

I ran into the same issue as the users in #429 and #1713

So I figured I make the information a bit more explicit. Not sure why the binary wasn't found automatically in the first place, but this helps :)

### DIFF
--- a/src/content/docs/reference/vscode.mdx
+++ b/src/content/docs/reference/vscode.mdx
@@ -52,7 +52,7 @@ The extension automatically loads the `biome.json` file from the workspace’s r
 
 The extension tries to use Biome from your project's local dependencies (`node_modules/@biomejs/biome`). We recommend adding Biome as a project dependency to ensure that NPM scripts and the extension use the same Biome version.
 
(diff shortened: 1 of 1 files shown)


ematipico:
That's incorrect, that's not the real binary. That's a JS file that forwards the CLI arguments to the real binary package, which depends on the OS and architecture. They are usually packages that start `@biomejs/cli-*`

oliversalzburg:
> That's incorrect, that's not the real binary. That's a JS file that forwards the CLI arguments to the real binary package, which depends on the OS and architecture. They are usually packages that start `@biomejs/cli-*`

Interesting. That makes sense. It _did_ resolve my integration issues though.

I'd be happy to aim for a better solution. Any suggestions?

ematipico:
cc @nhedger , he probably knows better

nhedger:
I wouldn't recommend pointing to that file, either. As mentioned, it's a shim, and it may not always work as expected, as seen in various issues. The extension should be able to detect the correct path itself, but if that's not the case in your situation, please open an issue with a reproduction, and we'll try to troubleshoot.

PR #1811, opened 2025-03-08
feat(playground): add language selector for single-file mode

siketyan:
## Summary

In single-file mode, it was not able to change the file language, especially for non JavaScript-like files (JSON, CSS, etc.). I added a language selector to change the file language quickly. Defaults to `TSX`.

image

### DIFF
--- a/src/playground/tabs/SettingsTab.tsx
+++ b/src/playground/tabs/SettingsTab.tsx
@@ -3,6 +3,7 @@ import {
 	AttributePosition,
 	Expand,
 	IndentStyle,
+	Language,
 	LintRules,
 	type PlaygroundState,
 	QuoteProperties,
@@ -16,9 +17,8 @@ import {
 	classnames,
 	createPlaygroundSettingsSetter,
 	getFileState,
-	isJsxFilename,
+	guessLanguage,
 	isScriptFilename,
-	isTypeScriptFilename,
 	modifyFilename,
 	normalizeFilename,
 } from "@/playground/utils";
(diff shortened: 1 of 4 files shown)


src/playground/tabs/SettingsTab.tsx
```diff
@@ -346,6 +360,38 @@ export default function SettingsTab({
...
+	setLanguage,
+}: { language: Language; setLanguage: (language: Language) => void }) {
+	return (
+		<section>
+			<div className="field-row">
+				<label htmlFor="language">Language</label>
+				<select
+					id="language"
+					name="language"
+					value={language ?? Language.TSX}
```
ematipico, line 374: Not a blocker, I wonder if there's a way to enforce a compiler failure in case we add new languages. I am concerned that once we add a new variant to `Language`, we forget to update this

ematipico (approved):
I tested it on mobile, and it works wonders! I left a couple of comments, but they aren't blockers

src/playground/tabs/SettingsTab.tsx
```diff
@@ -260,6 +260,18 @@ export default function SettingsTab({
...
+		Object.values(Language).find((ext) => ext === currentFileExtension) ??
+		Language.JS;
+
+	function setLanguage(language: Language): void {
+		renameFile(
+			currentFile,
+			`${currentFile.substring(0, currentFile.lastIndexOf("."))}.${language}`,
+		);
+	}
+
```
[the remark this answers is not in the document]
siketyan: I refactored to use `guessLanguage` (I added in this PR) and `modifyFilename` here. If we introduce `.d.ts` support, we only need to change the logic in the `modifyFilename` function

src/playground/tabs/SettingsTab.tsx
siketyan: I think adding a new component that maps a enum to a string text will help in this case. It will not be a straightforward representation for HTML, but missing case will never happen.

PR #1815, opened 2025-03-08
fix(playground): use generated `RuleDomains` type to fix typing issues

siketyan:
## Summary

Synced to pull the changes from biomejs/biome#5306 and biomejs/biome#5308, then fixed typing issues using the generated `RuleDomains` type. Plus, I added a `tsc` step to the workflow to ensure the code is valid in TypeScript on every PR.

### DIFF
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -12,7 +12,20 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
       - name: Setup Biome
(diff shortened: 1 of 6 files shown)


src/content/docs/linter/domains.mdx
siketyan, line 1: @ematipico Should we keep the original order of domains? If so, I will fix the `Ord` / `PartialOrd` impls to keep alphabetical order

ematipico: We should fix it and rely on `Ord` if possible, honestly I don't mind the kind of order as long as it's always the same

ematipico (approved):

src/content/docs/linter/domains.mdx
siketyan: Actually I don't mind too 😄 
I added impls to use alphabetical order: https://github.com/biomejs/biome/pull/5311

PR #1846, opened 2025-03-12
Beta blog post

arendjr:
## Summary

Also added plugin documentation, partially as a teaser for people to try out the beta.

Please make sure I didn't forget anything important!

~~TODO: @dyc3 Could you write a small piece about the HTML formatter? I'm less sure about the state it is in, or what people need to do to try it out.~~ **Update:** Done!

### DIFF
--- a/astro.config.ts
+++ b/astro.config.ts
@@ -257,6 +257,12 @@ export default defineConfig({
 										uk: "Сортування імпортів",
 									},
 								},
+								{
+									label: "Plugins",
+									link: "/analyzer/plugins",
+									badge: "beta",
+									translations: {},
+								},
 							],
 							translations: {
 								fr: "Analyseur",
@@ -415,6 +421,10 @@ export default defineConfig({
 								"pt-BR": "Extensão do Zed",
 								uk: "Розширення Zed",
 							},
+						},
+						{
+							label: "GritQL",
+							link: "/reference/gritql",
 						},
 					],
 				},
--- a/astro.config.ts
+++ b/astro.config.ts
@@ -28,6 +28,11 @@ const plugins = [
 				picture: "https://avatars.githubusercontent.com/u/2358560?s=96&v=4",
 				url: "https://bsky.app/profile/conaclos.bsky.social",
 			},
+			dyc3: {
+				name: "Carson McManus",
+				picture: "https://avatars.githubusercontent.com/u/1808807?v=4",
+				url: "https://github.com/dyc3",
+			},
 			ema: {
 				name: "Emanuele Stoppa",
 				picture: "https://avatars.githubusercontent.com/u/602478?v=4",
--- a/astro.config.ts
+++ b/astro.config.ts
@@ -424,10 +424,6 @@ export default defineConfig({
 						{
 							label: "GritQL",
 							link: "/reference/gritql",
-<<<<<<< HEAD
-=======
-							badge: "experimental",
->>>>>>> 76bafdd5 (fix: revert GritQL doc removal as it's already available in v1.9.4 (#1861))
 						},
 					],
 				},
--- a/astro.config.ts
+++ b/astro.config.ts
@@ -424,6 +424,10 @@ export default defineConfig({
 						{
 							label: "GritQL",
 							link: "/reference/gritql",
+<<<<<<< HEAD
+=======
+							badge: "experimental",
+>>>>>>> 76bafdd5 (fix: revert GritQL doc removal as it's already available in v1.9.4 (#1861))
 						},
 					],
 				},
--- a/astro.config.ts
+++ b/astro.config.ts
@@ -261,7 +261,6 @@ export default defineConfig({
 									label: "Plugins",
 									link: "/analyzer/plugins",
 									badge: "beta",
-									translations: {},
 								},
 							],
 							translations: {
--- a/src/content/docs/analyzer/plugins.mdx
+++ b/src/content/docs/analyzer/plugins.mdx
@@ -27,7 +27,7 @@ You can put a GritQL snippet in a file anywhere in your project, but be mindful
 you use the `.grit` extension. Then, you can simply enable it as a plugin with
 the following configuration:
 
-```
+```json name="biome.json"
 {
     "plugins": ["./path-to-plugin.grit"]
 }
--- a/src/content/docs/analyzer/plugins.mdx
+++ b/src/content/docs/analyzer/plugins.mdx
@@ -40,7 +40,7 @@ can see its results when running `biome lint` or `biome check`. For example:
 $ biome lint
 /packages/tailwindcss-config-analyzer/src/introspect.ts:12:17 plugin ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 
-  ℹ Prefer object spread instead of `Object.assign()`
+  ✖ Prefer object spread instead of `Object.assign()`
 
     10 │ function createContextFromConfig(config: Partial<Config>) {
     11 │        return createContext(
@@ -85,8 +85,10 @@ Biome currently supports one extra function:
 
 Registers a diagnostic to be reported whenever the pattern matches.
 
-Requires two arguments:
+Supports three arguments:
 
-- `span`: The syntax node to attach the diagnostic to. This is typically a
-  variable that you matched within a code snippet.
-- `message`: The message to show with the diagnostic.
+- `span` (required): The syntax node to attach the diagnostic to. This is
+  typically a variable that you matched within a code snippet.
+- `message` (required): The message to show with the diagnostic.
+- `severity`: The severity of the diagnostic. Allowed values are: `hint`,
+  `information`, `warning`, and `error`. By default, `error` is used.
--- a/src/content/docs/analyzer/plugins.mdx
+++ b/src/content/docs/analyzer/plugins.mdx
@@ -1,0 +1,52 @@
+---
+title: Analyzer Plugins
+description: An overview of how to use Biome's Analyzer Plugins
+---
+
+:::note
+Plugins are currently only available in the [2.0 beta release](/blog/biome-v2-0-beta).
+:::
+
+Biome Analyzer supports [GritQL](/reference/gritql/) plugins. Currently, these
+plugins allow you to match specific code patterns and register customized
+diagnostic messages for them.
+
+An example of a plugin is:
+
+```grit
+`$fn($args)` where {
+    $fn <: `Object.assign`,
+    register_diagnostic(
+        span = $fn,
+        message = "Prefer object spread instead of `Object.assign()`"
+    )
+}
+```
+
+You can put a GritQL snippet in a file anywhere in your project, but be mindful
+you use the `.grit` extension. Then, you can simply enable it as a plugin with
+the following configuration:
+
+```
+{
+    "plugins": ["./path-to-plugin.grit"]
+}
+```
+
+The plugin will now be enabled on all files the analyzer runs on.
+
+## Plugin API
+
+In addition to Grit's
+[built-in functions](https://docs.grit.io/language/functions#built-in-functions),
+Biome currently supports one extra function:
+
+### `register_diagnostic()`
+
+Registers a diagnostic to be reported whenever the pattern matches.
+
+Requires two arguments:
+
+- `span`: The syntax node to attach the diagnostic to. This is typically a
+  variable that you matched within a code snippet.
+- `message`: The message to show with the diagnostic.
--- a/src/content/docs/analyzer/plugins.mdx
+++ b/src/content/docs/analyzer/plugins.mdx
@@ -11,7 +11,7 @@ Biome Analyzer supports [GritQL](/reference/gritql/) plugins. Currently, these
 plugins allow you to match specific code patterns and register customized
 diagnostic messages for them.
 
-An example of a plugin is:
+Here is an example of a plugin that reports on all usages of `Object.assign()`:
 
 ```grit
 `$fn($args)` where {
@@ -33,7 +33,47 @@ the following configuration:
 }
 ```
 
-The plugin will now be enabled on all files the analyzer runs on.
+The plugin will now be enabled on all supported files the analyzer runs on. You
+can see its results when running `biome lint` or `biome check`. For example:
+
+```shell
+$ biome lint
+/packages/tailwindcss-config-analyzer/src/introspect.ts:12:17 plugin ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
+
+  ℹ Prefer object spread instead of `Object.assign()`
+
+    10 │ function createContextFromConfig(config: Partial<Config>) {
+    11 │        return createContext(
+  > 12 │                resolveConfig(Object.assign({}, DEFAULT_CONFIG, config)),
+       │                              ^^^^^^^^^^^^^
+    13 │        );
+    14 │ }
+```
+
+## Target Languages
+
+A GritQL snippet always attempts to match against a given _target language_.
+If no target language is specified, JavaScript or one of its super languages is
+assumed.
+
+If you want to use a different target language, you must specify it explicitly.
+For example, here is a CSS plugin to report any selector that sets a color
+outside the allowed `.color-*` classes:
+
+```grit
+language css;
+
+`$selector { $props }` where {
+    $props <: contains `color: $color` as $rule,
+    not $selector <: r"\.color-.*",
+    register_diagnostic(
+        span = $rule,
+        message = "Don't set explicit colors. Use `.color-*` classes instead."
+    )
+}
+```
+
+We currently do not support other target languages than JavaScript and CSS.
 
 ## Plugin API
 
--- a/src/content/docs/blog/biome-v2-0-beta.md
+++ b/src/content/docs/blog/biome-v2-0-beta.md
@@ -7,6 +7,7 @@ summary: |
   release can be as smooth as possible.
 authors:
   - arendjr
+  - dyc3
   - team
 date: 2025-03-31
 cover:
@@ -22,7 +23,7 @@ If you would like to try it out, you can update Biome and migrate your configura
 
 ```shell
 npm install --save-dev --save-exact @biomejs/biome@beta
-npx @biomejs/biome migrate
+npx @biomejs/biome@beta migrate
 ```
 
 ## New features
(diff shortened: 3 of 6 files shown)


ematipico:
@arendjr what's the objective of this blog post? I would expect a blog post even when we officially release v2, so I wonder what are the differences and what this blog should be about

astro.config.ts
```diff
@@ -257,6 +257,12 @@ export default defineConfig({
 										uk: "Сортування імпортів",
 									},
 								},
+								{
+									label: "Plugins",
+									link: "/analyzer/plugins",
+									badge: "beta",
+									translations: {},
```
ematipico, line 264:
```suggestion
```

src/content/docs/analyzer/plugins.mdx
```diff
@@ -0,0 +1,52 @@
...
+
+:::note
+Plugins are currently only available in the [2.0 beta release](/blog/biome-v2-0-beta).
+:::
+
+Biome Analyzer supports [GritQL](/reference/gritql/) plugins. Currently, these
+plugins allow you to match specific code patterns and register customized
+diagnostic messages for them.
+
+An example of a plugin is:
```
ematipico, line 14: Let's explain what this plugin does in this line.

src/content/docs/analyzer/plugins.mdx
```diff
@@ -0,0 +1,52 @@
...
+
+You can put a GritQL snippet in a file anywhere in your project, but be mindful
+you use the `.grit` extension. Then, you can simply enable it as a plugin with
+the following configuration:
+
+```
+{
+    "plugins": ["./path-to-plugin.grit"]
+}
+```
```
ematipico, line 34:
``````suggestion
```json name="biome.json"
{
    "plugins": ["./path-to-plugin.grit"]
}
```
``````

Also, users might want to publish their plugins as npm packages, so we should provide instructions on how they do so and how they can import them. If they can't, we should say it somewhere.

File: src/content/docs/analyzer/plugins.mdx [hunk 4]
```diff
@@ -0,0 +1,52 @@
...
+
+The plugin will now be enabled on all files the analyzer runs on.
```
ematipico, line 36: This assumes that plugins runs in all languages supported by Biome. Is that the case? If so, we should specify it and maybe add a link to `/reference/language-support`

src/content/docs/analyzer/plugins.mdx
```diff
@@ -0,0 +1,52 @@
...
+
+## Plugin API
```
ematipico, line 38: Before this header, we should:
- show the command to run
- show a diagnostic printed by this plugin

I know plugins have their diagnostic category now, so we should highlight that part too

src/content/docs/analyzer/plugins.mdx
```diff
@@ -0,0 +1,52 @@
...
+
+In addition to Grit's
+[built-in functions](https://docs.grit.io/language/functions#built-in-functions),
+Biome currently supports one extra function:
+
+### `register_diagnostic()`
+
+Registers a diagnostic to be reported whenever the pattern matches.
```
ematipico, line 46: The API doesn't mention the diagnostic category emitted by the plugin (error, warn, info). You should mention it and explain if users can control it and, if so, how. (or if they can't)

src/content/docs/blog/biome-v2-0-beta.md
```diff
@@ -0,0 +1,177 @@
+---
+title: Biome v2.0 beta
+description: A major milestone in the making
+summary: |
+  Biome 2.0 will be packed with major features and many smaller fixes, rules, and other improvements.
+  There's a lot to unpack, and we request the community's help testing this beta, so the final
+  release can be as smooth as possible.
+authors:
+  - arendjr
+  - team
```
ematipico, line 10: Remember to add `dyc3` once they add their part of the article :)

src/content/docs/blog/biome-v2-0-beta.md
```diff
@@ -0,0 +1,177 @@
...
+socialImage: "@/assets/social-logo.png"
+---
+
+After hard work from our team, Biome's long-awaited 2.0 release is nearing completion. It will be packed with many large features, so we would like your help testing it with a public beta!
+
+If you would like to try it out, you can update Biome and migrate your configuration using the following commands:
+
+```shell
+npm install --save-dev --save-exact @biomejs/biome@beta
+npx @biomejs/biome migrate
```
ematipico, line 25:
```suggestion
npx @biomejs/biome@beta migrate
```

`npx` downloads packages straight from the registry if I remember correctly

src/content/docs/blog/biome-v2-0-beta.md
```diff
@@ -0,0 +1,177 @@
...
+```
+
+## New features
+
+While the final 2.0 release may still have small changes in its final feature set, here's what you can expect in the beta:
+
+- **Plugins:** You can write custom lint rules using GritQL.
+- **Domains:** Domains help to group lint rules by technology, framework, or well, domain. Thanks to domains, your default set of recommended lint rules will only include those that are relevant to your project.
```
ematipico, line 33:
```suggestion
- **Linter domains:** Domains help to group lint rules by technology, framework, or well, domain. Thanks to domains, your default set of recommended lint rules will only include those that are relevant to your project.
```

Just a suggestion to frame the name "domain" in the right context

src/content/docs/blog/biome-v2-0-beta.md
```diff
@@ -0,0 +1,177 @@
...
+  - `noInvalidNewBuiltin`
+  - `noNewSymbol`
+  - `useShorthandArrayType`
+  - `useSingleCaseStatement`
+- **BREAKING:** Many deprecated options, including some from the Rome days, have been removed.
+- Added a new option `javascript.parser.jsxEverywhere` to control whether Biome should expect JSX syntax in `.js`/`.mjs`/`.cjs` files.
+- Improved monorepo support: The rule [`noUndeclaredDependencies`](https://biomejs.dev/linter/rules/no-undeclared-dependencies/) now works correctly in monorepos by using the nearest `package.json` file, instead of only the root one.
+- We have enabled support for `.editorconfig` files by default.
+- Changed default formatting of `package.json` to align better with formatting by package managers.
+
```
ematipico, line 174: It may be worth mentioning that we removed all legacy stuff related to Rome: `rome.json`, rome suppression comments, environment variables, etc.

src/content/docs/blog/biome-v2-0-beta.md
```diff
@@ -0,0 +1,177 @@
...
+### And more!
+
+For the full list of changes, please refer to our [changelog](/internals/changelog/).
```
ematipico, line 177: Unless we are able to provide a changelog for the beta, we should remove it. We still don't know yet **how** we will generate the beta, and the changelog creation is now owned by changesets

src/content/docs/reference/gritql.mdx
```diff
@@ -11,11 +11,11 @@ nesting, and variables.
 
 GritQL is [open-source](https://github.com/getgrit/gritql/) and created by [Grit.io](https://grit.io/).
 
-Biome is integrating GritQL for two purposes:
+Biome uses GritQL for two purposes:
 
+- Our [Analyzer Plugins](/analyzer/plugins).
```
ematipico, line 16:
```suggestion
- The [Analyzer Plugins](/analyzer/plugins).
```

Let's avoid "we" unless we speak about the team

astro.config.ts
ematipico, line 264:
```suggestion
```
Not needed for now

ematipico (commented):
Left some comments and questions to improve the plugins part page. Let me know what you think

src/content/docs/analyzer/plugins.mdx
arendjr: Yeah, good point. I haven't really tested if it works with NPM packages yet. Maybe something to clear up before the final 2.0 release.

src/content/docs/analyzer/plugins.mdx
arendjr: Excellent point! Currently it's only CSS  and JS only indeed.

src/content/docs/blog/biome-v2-0-beta.md
arendjr: Yeah, I had assumed we could create a new changelog from the changesets, but let's see how that works out :sweat_smile:

ematipico: It is possible, but we need to change the script that does that

src/content/docs/blog/biome-v2-0-beta.md
arendjr: It's there a few lines up, though maybe it was easy to miss:

> Many deprecated options, including some from the Rome days, have been removed.

I'll reword it to "including some that still referenced the old Rome name" to be less obscure.

src/content/docs/analyzer/plugins.mdx
arendjr: Huh, it always uses the default severity... which I guess is error? That might be a nice thing to make configurable before 2.0 still :)

src/content/docs/blog/biome-v2-0-beta.md
arendjr: Only if the package is not installed, I thought, which is the confusing (and rather insecure) thing about `npx`: it is used for both local and remote command invocations. You make one typo trying to execute a local command, and you'll be downloading and executing code off the internet :sweat:

src/content/docs/blog/biome-v2-0-beta.md
arendjr: I think I'll leave this as is. The context is clear right after, and "Domains" is much stronger terminology than "Linter domains", especially since there's no real room for confusion.

arendjr:
@ematipico The purpose of the blog post is the same as with any release, to draw awareness, and specifically in this case: to get testers.

For 2.0 we will have another post, and it may look mostly the same as this one. That's should be fine, for instance TypeScript announcements are the same: the final release announcements are mostly a repeat of the release candidates and/or beta before them. Of course the introductory text may still differ, and smaller differences may exist too.

src/content/docs/analyzer/plugins.mdx
arendjr: Feature added: https://github.com/biomejs/biome/pull/5363
And documentation added on the page here.

src/content/docs/blog/biome-v2-0-beta.md
```diff
@@ -0,0 +1,178 @@
...
+
+This way, Biome's total set of recommended rules should be most relevant to your specific project needs.
+
+And finally, domains can add global variables to the `javascript.globals` setting. This should make Biome even easier to setup.
+
+### Multi-file analysis
+
+Before version 2.0, Biome lint rules could only operate on one file at a time. This brought us far, but many of the more interesting rules require information from other files too.
+
+To accomplish this, we have added a _file scanner_ to Biome that scans all the files in your project and indexes them, similar to what an LSP service might do in your IDE. We're not going to beat around the bush: Scanning projects means that Biome has become slower for many projects. But we do believe the ability to do multi-file analysis is worth it. And without a scanner, multi-file analysis would become _even slower_, as rules would need to perform ad-hoc file system access individually.
```
Conaclos, line 98: > Scanning projects means that Biome has become slower for many projects

Have we some data about that? Just curious how slower is it.

src/content/docs/blog/biome-v2-0-beta.md
```diff
@@ -0,0 +1,178 @@
...
+- [`noUnwantedPolyfillio`](https://biomejs.dev/linter/rules/no-unwanted-polyfillio)
+- [`useConsistentObjectDefinition`](https://biomejs.dev/linter/rules/use-consistent-object-definition/)
+- [`useForComponent`](https://biomejs.dev/linter/rules/use-for-component/)
+
+### Miscellaneous
+
+- **BREAKING:** The configuration fields `include` and `ignore` have been replaced with a single `includes` field.
+- **BREAKING:** Reworked some recommended rules recommended to be less pedantic and blocking. This is a breaking change if your project relied on those rules to block the CI in case of violations. If you used the `migrate` command, the behaviour should remain as before.
+- **BREAKING:** The `style` rules aren't recommended anymore. If you used the `migrate` command, the behaviour should remain as before.
+- **BREAKING:** Cleaned up deprecated rules:
```
Conaclos, line 164:
```suggestion
- **BREAKING:** Removed deprecated rules:
```

src/content/docs/blog/biome-v2-0-beta.md
arendjr: I’m afraid it’s very project-dependent. In Biome’s own project, I think it added 9 seconds, which is pretty bad until you realise it’s scanning mostly Rust stuff. That’s… still bad, but also means we may have room to be smarter about scanning.

On pure JS projects, it seems the overhead is much less, but it’ll still depend on the project of course.

As you mentioned this, I think the beta is also a good opportunity to ask people’s experiences. There’s still time to change the scanner behaviour for 2.0 if people complain. I’ll reword this a bit to make it more of call-to-action if people are disappointed with the overhead.

Conaclos: > it’s scanning mostly Rust stuff.

You mean traversing unnecessary folders?
How is this impacting the scanner perf?

Dummy question: have we taken a look at how OXC achieves its multi-file analysis?

arendjr: > You mean traversing unnecessary folders?
> How is this impacting the scanner perf?

Yes indeed, but I haven't actually measured what the impact of it is the whole.

> Dummy question: have we taken a look at how OXC achieves its multi-file analysis?

Not a dumb question at all! I was almost going to say: OXC doesn't support multi-file analysis yet, but the truth is a bit more nuanced than that. They do have limited multi-file analysis, which is built directly upon the module graph that they build using `oxc-resolver`. By their own admission however, this is more of a proof-of-concept and beyond their `no-cycles` rule, it doesn't really appear to be used in the linter.

When I updated `oxc-resolver` for our use cases, I kinda started to assume they didn't use `oxc-resolver` for multi-file analysis yet. The reason for this is because it isn't designed with cache invalidation in mind. I think now, but this is a presumption on my part, that they're designing these tools with a CLI focus first and foremost. If you only focus on the CLI use case, you don't really need cache invalidation: You can just build a module graph on demand, and assume it'll remain valid until the end of your session.

But for our multi-file analysis I've explicitly taken the LSP use case into account, and it lead me into a wholly different design direction. In my original [multi-file analysis RFC](https://github.com/biomejs/biome/discussions/4653) I basically contrasted the approach that we're taking (scanner + watcher, with papaya-based hash maps that can easily invalidate segments of the data we use for multi-file analysis) with the approach taken by Astral, where they use `salsa` which allows for on-demand graph creation with very precise invalidation. Oxc's approach is closer to `salsa`... but without the invalidation. That's fine if you're building a CLI tool only, but for something that requires first-class IDE support, I think they have a long ways to go still.

But yeah, OXC is certainly faster :)

> Also, does this impact the perf of only applying the formatter? i.e. biome format.

It does, yes. The main reason is because we want to use the scanner for our monorepo support too. We need to scan the project layout and be able to find nested `biome.json` files, to be able to determine the correct formatter settings inside every nested project. Even if it's not for monorepo support, people may also be making use of the daemon. And the daemon is horribly inefficient in its current design, where the client does all the file traversal and then sends each document individually to the daemon. We should move away from that, but that also implies a higher reliance on the scanner.

Of course, neither of these reasons really apply to how we do things _yet_, but it felt unnecessary to carve out a temporary exception for commands such as `biome format`.

siketyan: JFYI, I am dog-fooding the preview build of Biome v2 in my projects, and experiencing slow down from 3 seconds to 24 seconds in a large project (approx. 5500 files)

arendjr: @siketyan Awesome! Not the slowdown, of course, but that's exactly the kind of feedback we need :) Is it an open-source repository that I could have a peek at?

siketyan: @arendjr
Actually it's a repository under my company so I cannot share it to you, sorry.
I can check where's the hotspot of the executin time if you're interested :)

siketyan: It was an easy hotspot, I dropped a fix 😄 
And it was not from multi-file scanning, sorry about that.
https://github.com/biomejs/biome/pull/5380

arendjr: With @siketyan 's fix, running Biome on Biome's own repository is also only 2 seconds, so even with a lot of Rust code in the repo it doesn't seem a big issue anymore :)

src/content/docs/blog/biome-v2-0-beta.md
```diff
@@ -0,0 +1,222 @@
...
+### `noFloatingPromises`
+
+With Biome's linter we have always strived to provide a battery-included approach to linting. This means we're not just aiming to replace ESLint, but also its plugins. One of the hardest plugins to replace is **`typescript-eslint`**.
+
+Biome has featured some rules from `typescript-eslint` for a while now, but we could never replace all rules, because they relied on type information for their analysis. And in order to get type information, `typescript-eslint` relies on `tsc` itself, which is rather slow and also complicates setup.
+
+This is about to change. With Biome 2.0, we're introducing a first version of the [`noFloatingPromises`](https://next.biomejs.dev/linter/rules/no-floating-promises) rule, one of the most-requested rules that relies on type information. In fairness, we should not consider it more than a proof-of-concept right now, because there are some notable limitations to its capabilities:
+
+* It doesn't understand complex types yet.
+* It cannot yet do type inference yet.
```
ematipico, line 121:
```suggestion
* It cannot do type inference yet.
```

src/content/docs/blog/biome-v2-0-beta.md
```diff
@@ -0,0 +1,222 @@
...
+* It can currently only analyse types that occur in the same file.
+
+Still, its capabilities are sufficient to catch some of the low-hanging fruit. Consider this small snippet:
+
+```ts
```
ematipico, line 126:
``````suggestion
```js name="example.js"
``````

This is a great example because TypeScript is not even needed :)

File: src/content/docs/blog/biome-v2-0-beta.md [hunk 17]
```diff
@@ -0,0 +1,222 @@
...
+async function returnsPromise() { /* ... */ }
+
+returnsPromise().then(() => {});
+```
+
+It will trigger the following diagnostic:
+
+```
+example.ts:3:1 lint/nursery/noFloatingPromises ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
```
ematipico, line 135:
```suggestion
example.js:3:1 lint/nursery/noFloatingPromises ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
```

ematipico:
@arendjr  I added a banner to the homepage [`ccc1a14` (#1846)](https://github.com/biomejs/website/pull/1846/commits/ccc1a1439bfa89ea6e17cbbce40df5380a3badb2)

ematipico (approved):

PR #1858, opened 2025-03-14
chore: revert and pin to v1.9.4

siketyan:
## Summary

The current `main` branch contains some changes from v2, and the users can get incorrect information as the latest released version is v1.9.4. I pinned the source of codegen to v1.9.4, re-generate all documentation, and changed the synchronise workflow to do stuff in the `next` branch.

### DIFF
--- a/codegen/src/rules_sources.rs
+++ b/codegen/src/rules_sources.rs
@@ -71,9 +71,7 @@ description: A page that maps lint rules from other sources to Biome
                             biome_link: format!("/linter/rules/{kebab_rule_name}"),
                             source_link: source.to_rule_url(),
                             source_rule_name: source.as_rule_name().to_string(),
-                            inspired: metadata
-                                .source_kind
-                                .map_or(false, |kind| kind.is_inspired()),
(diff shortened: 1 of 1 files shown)


ematipico (approved):
Wow, you thought about everything! Thank you ❤️

astro.config.ts
```diff
@@ -416,11 +416,6 @@ export default defineConfig({
 								uk: "Розширення Zed",
 							},
 						},
-						{
-							label: "GritQL",
-							link: "/reference/gritql",
-							badge: "experimental",
-						},
```
arendjr, line 423: Why was this removed, along with the entire `gritql.mdx` file? That was already part of 1.9.

siketyan: Sorry, was thought it is not available yet. I'll revert it asap

siketyan: Reverted. Thank you for noticing my mistake 🙇

arendjr: No worries! Thanks for the effort!

PR #1865, opened 2025-03-15
feat: add version selector as the link to next.biomejs.dev

siketyan:
## Summary

Add a version selector for redirecting to the v2.0 beta documentation.
Not sure this is the right way in Astro + Starlight, I added the selector into the `LanguageSelect` component as `starlight-blog` does for [`ThemeSelect`](https://github.com/HiDeoo/starlight-blog/blob/3f9e45506b240a6c837e0f601b9cf865ca6473e1/packages/starlight-blog/overrides/ThemeSelect.astro).

### DIFF
--- a/src/components/starlight/LanguageSelect.astro
+++ b/src/components/starlight/LanguageSelect.astro
@@ -1,0 +1,43 @@
+---
+import Default from "@astrojs/starlight/components/LanguageSelect.astro";
+import Select from "@astrojs/starlight/components/Select.astro";
+---
+
+<Default><slot /></Default>
+<version-select>
+    <Select
+        icon="biome"
+        label="Version"
+        value="v1"
+        width="fit-content"
+        options={[
+            { label: "v1.x", value: "v1", selected: true },
+            { label: "beta", value: "v2", selected: false },
+        ]}
+    />
+</version-select>
+
+<script>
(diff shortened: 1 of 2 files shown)


src/components/starlight/LanguageSelect.astro
```diff
@@ -0,0 +1,43 @@
...
+import Default from "@astrojs/starlight/components/LanguageSelect.astro";
+import Select from "@astrojs/starlight/components/Select.astro";
+---
+
+<Default><slot /></Default>
+<version-select>
+    <Select
+        icon="biome"
+        label="Version"
+        value="v1"
```
siketyan, line 11: We'll update this to `v2` after merged into the `next` branch.

src/components/starlight/LanguageSelect.astro
dyc3, line 1: nit: Shouldn't this be named `VersionSelect.astro`?

siketyan: @dyc3 I admit it's a bit confusing. This component LanguageSelect extends the original LanguageSelect to add a custom version selector next to the language selector. It's hacky as Starlight doesn't have any slots to add custom components here

arendjr (approved):

PR #1867, opened 2025-03-15
Document `includes` behaviour

arendjr:
## Summary

Updated both the reference and the configuration guide to update the docs on how `includes` works.

I've removed the `organizeImports` section, because it was mostly about the old `include` and `ignore`, but I didn't yet add a new reference on the `assist` options.

### DIFF
--- a/src/content/docs/guides/configure-biome.mdx
+++ b/src/content/docs/guides/configure-biome.mdx
@@ -68,7 +68,7 @@ We disabled the formatter for JSON files.
 ```
 
 :::note
-Biome refers as `javascript` all variants of the JavaScript language.
+Biome refers to all variants of the JavaScript language as `javascript`.
 This includes TypeScript, JSX and TSX.
 :::
 
@@ -101,21 +101,12 @@ Here's an example:
 will use the configuration file `app/frontend/biome.json`;
 
 :::caution
-Biome doesn't support nested `biome.json` files, neither in CLI nor in LSP. [Follow and help the support in the related issue](https://github.com/biomejs/biome/issues/2228)
-:::
-
-:::note
-Biome commands support using the `--config-path` option and the `BIOME_CONFIG_PATH` environment variable.
-This allows you to specify a custom configuration file or a directory where Biome can look for a `biome.json` or `biome.jsonc` file.
-When you use `--config-path` or `BIOME_CONFIG_PATH`, the standard configuration file resolution process is **disabled**.
-
-If `--config-path` or `BIOME_CONFIG_PATH` points directly to a file, you can use names other than `biome.json` or `biome.jsonc`.
-Biome will read a `.json` file using a standard JSON parser.
-For files with other extensions, Biome will treat them as `.jsonc` files, using a more flexible JSON parser that allows comments and trailing commas.
+Biome doesn't yet support nested `biome.json` files, neither in CLI nor in LSP.
+[Follow and help the support in the related issue](https://github.com/biomejs/biome/issues/2228)
 :::
 
 
-## Share a configuration file
+## Sharing a configuration file
 
 The `extends` field allows you to split your configuration across multiple files.
 This way, you can share common settings across different projects or folders.
@@ -170,11 +161,11 @@ To avoid a breaking change with how the existing resolution works, paths startin
 For more information about the resolution algorithm, refer to the [Node.js documentation](https://nodejs.org/api/esm.html#resolution-and-loading-algorithm).
 
 
-## Ignore files
+## Specifying files to process
 
-The first way to control which files and directories are processed by Biome is to list them in the CLI.
-In the following command, we format only `file1.js` and all the files in the `src` directory.
-The directories are recursively traversed.
+The first way to control which files and folders are processed by Biome is to
+list them in the CLI. In the following command, we only format `file1.js` and
+all the files in the `src` folder, because folders are recursively traversed.
 
 ```shell
 biome format file1.js src/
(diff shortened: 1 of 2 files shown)


chansuke (approved):

src/content/docs/guides/configure-biome.mdx
```diff
@@ -187,33 +178,36 @@ Some shells don't support the recursive glob `**`.
...
+here are processed by the linter, the formatter and the assist, unless specified
+otherwise. For the individual tools, you can further refine the matching files
+using `<tool>.includes`.
+
+:::important
+The global `files.includes` has slightly different semantics than the more
+specific `<tool>.includes` fields. Any file or folder that doesn't match
+`files.includes` is excluded from use by any of Biome's tools. This means that
+any tool-specific `includes` field can never match a file that doesn't also
+match `files.includes`.
```
Conaclos, line 199: Actually it is even a bit more subtle.
Folders are only matched against negated patterns because folders like `src` doesn't match a pattern like `src/file.js`.

```suggestion
specific `<tool>.includes` fields. Any file that doesn't match
`files.includes` is excluded from use by any of Biome's tools. This means that
any tool-specific `includes` field can never match a file that doesn't also
match `files.includes`.
Also, Biome will not traverse a folder that matches a negated pattern from `files.includes`.
```

arendjr: Thanks! I don't think I would change the description though: After all, _not traversing_ a folder causes all the files in that folder to not match. So the original wording already implies the explanation you're giving. But while your explanation does help to explain _why_ those files don't match, it is actually harder to understand for the reader because you're overwhelming them with technical details that don't really matter for the end result.

src/content/docs/guides/configure-biome.mdx
```diff
@@ -187,33 +178,36 @@ Some shells don't support the recursive glob `**`.
...
+
+Paths and globs inside Biome's configuration file are always resolved relative
+to the folder the configuration file is in.
+
+`files.includes` applies to all of Biome's tools, meaning the files specified
+here are processed by the linter, the formatter and the assist, unless specified
+otherwise. For the individual tools, you can further refine the matching files
+using `<tool>.includes`.
+
+:::important
```
ematipico, line 194: This isn't part of [starlight](https://starlight.astro.build/guides/authoring-content/#asides). Did you want to use `:::note` or `:::caution`?

src/content/docs/guides/configure-biome.mdx
ematipico, line 199: It seems that this paragraph repeats what was stated before, just in different words. 

> `files.includes` applies to all of Biome's tools, meaning the files specified
here are processed by the linter, the formatter and the assist, unless specified
otherwise. For the individual tools, you can further refine the matching files
using `<tool>.includes`.

To me this phrase is clear enough, so maybe this paragraph should talk about `<tool>.includes` instead?

src/content/docs/guides/configure-biome.mdx
```diff
@@ -224,68 +218,18 @@ And run the following command:
...
 
-If we run the following command, no files are linted because the `test` directory is explicitly ignored for the linter.
+If we run the following command, no files are linted because files inside the
+`test/` folder are explicitly ignored for the linter.
 
 ```shell
 biome lint test/
 ```
 
-Biome resolves the globs relatively from the working directory.
```
ematipico, line 236: The PR doesn't mention anything about the removal of this section. Is there a reason? 

Since we changed how we resolve globs and folders, maybe we should **update** this section accordingly.

src/content/docs/reference/configuration.mdx
```diff
@@ -29,71 +31,74 @@ If you have problems with resolving the physical file, you can use the one publi
...
+A list of paths to other Biome configuration files. Biome resolves and applies
+the configuration settings from the files contained in the `extends` list, and
+eventually applies the options contained in this `biome.json`/`biome.jsonc`
+file.
 
-## `files`
-
-### `files.maxSize`
+Files are provided in order from least relevant to most relevant, with the
+settings in the current file being leading.
```
ematipico, line 40: This phrase is a bit out of context. What "least relevant to most relevant" mean? You probably wanted to talk about the **order** of the items inside `extends`, is that right? If so, than we should mention it and maybe provide an example.

src/content/docs/reference/configuration.mdx
```diff
@@ -1109,30 +1109,16 @@ When a file is matched against an override pattern, the configuration specified
...
-
-A list of Unix shell style patterns. Biome will not apply the override to files that match the pattern.
+### `overrides.<ITEM>.includes`
 
+A list of [glob patterns](https://en.wikipedia.org/wiki/Glob_(programming)) of
+files for which to apply customised settings.
 
-
-```json title="biome.json"
+```json title="biome.jsonc"
```
ematipico, line 1117:
```suggestion
```jsonc title="biome.jsonc"
```

ematipico (commented):
Thank you! I left some suggestions

src/content/docs/guides/configure-biome.mdx
arendjr: Thanks, looks like I looked at a different markdown flavour reference 😅

src/content/docs/guides/configure-biome.mdx
arendjr: It was primarily meant to clarify what “refinement” means, since I felt it was far from obvious. The fact that tool-specific files cannot match things outside the global files bears some explicit calling out, I think.

src/content/docs/reference/configuration.mdx
arendjr: Yes indeed, I’ll try to reword this a bit.

src/content/docs/guides/configure-biome.mdx
arendjr: Yes, it was updated, but I see it was updated in the process :)

A little above it says:

> Paths and globs inside Biome's configuration file are always resolved relative to the folder the configuration file is in.

ematipico: Yes I read that part, but to be honest I don't like it. We just removed a good example with a small paragraph. It seems a step down from before

arendjr: Ah, I see. I honestly felt that the example was only necessary because the old behaviour was so unintuitive. Do we really still need an example to explain that Biome does what they already expect :)

ematipico: Regardless of how intuitive a feature is/was, we wrote how Biome functioned and also made an example of how it works, and that's how our users learnt the behaviour.

Now, the expectations might seem more intuitive, but it still regards some examples to explain those expectations, especially in projects where there might be multiple configuration files and the use of `extends`. As a result:
- beginners will learn from zero with examples, it can't hurt;
- new users will find it intuitive, but if something doesn't click, the example can help;
- old users will re-learn the concept with examples, and they quickly change their setup;

Providing examples is also useful for us when triaging issues: if the bug doesn't respect the documentation, then it's not a bug. However, if it is, it's also an opportunity to fix our documentation.

arendjr: Alright, I've added an updated example under the "Sharing a configuration file" section. While at it, I've also moved this section to the bottom of the guide, seeing how I do think it's a bit more of an advanced topic. I think the flow of the guide has improved a bit like this. The earlier line that said "Paths and globs inside Biome's configuration file are always resolved relative to the folder the configuration file is in." has been updated to mention there's an exception in the face of `extends` and links to the relevant section.

Also added a line to mention that `extends` doesn't work recursively.

ematipico (approved):

PR #1868, opened 2025-03-16
fix(playground): enable html formatter on the playground

siketyan:
## Summary

The html formatter has been disabled by default in https://github.com/biomejs/biome/pull/5354, we need to explicitly enable it in the playground.


arendjr (approved):
Thank you!

PR #1933, opened 2025-03-25
feat(codegen): remove "not released yet" notification in favour of "only available in beta"

siketyan:
## Summary

v2.0 beta is now available, the "not released yet" notification on some rules is confusing. Replaced it with a new "only available in beta" info-level notification.


ematipico:
We have to remember to switch it back before merging `next` to `main`

PR #2056, opened 2025-04-16
range suppressions now require a matching end

hanseltime:
## Summary

As part of https://github.com/biomejs/biome/pull/5619, we updated range-suppressions to require a matching `-end` so that suppressions can't leak and disable critical rules further down in the file.

Per that discussion, we need to update the documentation of range suppressions so that we don't make people think they can use open-ended `-start` suppressions.


hanseltime:
`dev` worked just fine locally but the `next` branch and this branch both fail to build due to the missing linting rule.  If someone knows why, feel free to let me know how to remedy it

ematipico (approved):

PR #2071, opened 2025-04-19
combine duplicated rule entries into a single row with multiple properties

haejunejung:
## Summary

Merged multiple rows for the same rule (`noIrregularWhitespace`) into a single entry by combining their language-specific icons under one property cell.


ematipico:
Thank you, but this is an auto generated file, as mentioned at the top of the file.

We fixed this in the upcoming version of the docs

PR #2136, opened 2025-04-30
feat(playground): add semantic model tab

dyc3:
## Summary

Expose a new tab for the semantic model debug printer. This will not work until https://github.com/biomejs/biome/pull/5821 is merged, and this PR is rebased to include it.


chansuke (approved):
I've read your merged [PR](https://github.com/biomejs/biome/pull/5821) as well and looks nice.

PR #2154, opened 2025-05-03
docs: add @vitest/eslint-plugin as rule source

drwpow:
## Summary

`@vitest/eslint-plugin` mirrors `eslint-plugin-jest`, so Biome can count it as a rule source! Add mentions of Vitest to docs everywhere Jest is.

CleanShot 2025-05-03 at 12 00 33@2x

> [!NOTE]
> There’s one Jest rule, `no-export`, which does NOT exist in the Vitest plugin (dunno why). So you’ll notice out of 7 Jest rules, only 6 rules added Vitest as a source. This is intentional.

Resolves part of https://github.com/biomejs/biome/discussions/5862.

### DIFF
--- a/src/content/docs/linter/rules-sources.mdx
+++ b/src/content/docs/linter/rules-sources.mdx
@@ -6,43 +6,43 @@
 title: Rules sources
 description: A page that maps lint rules from other sources to Biome
 ---
-    
+
 ## Biome exclusive rules
-- [noAccumulatingSpread](/linter/rules/no-accumulating-spread) 
-- [noConsoleLog](/linter/rules/no-console-log) 
-- [noConstEnum](/linter/rules/no-const-enum) 
-- [noDelete](/linter/rules/no-delete) 
-- [noDuplicateObjectKeys](/linter/rules/no-duplicate-object-keys) 
-- [noDynamicNamespaceImportAccess](/linter/rules/no-dynamic-namespace-import-access) 
-- [noEmptyTypeParameters](/linter/rules/no-empty-type-parameters) 
-- [noEnum](/linter/rules/no-enum) 
-- [noEvolvingTypes](/linter/rules/no-evolving-types) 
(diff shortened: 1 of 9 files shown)


src/content/docs/linter/rules-sources.mdx
```diff
@@ -6,43 +6,43 @@
...
-- [useImportExtensions](/linter/rules/use-import-extensions) 
-- [useNodeAssertStrict](/linter/rules/use-node-assert-strict) 
-- [useShorthandArrayType](/linter/rules/use-shorthand-array-type) 
-- [useSimpleNumberKeys](/linter/rules/use-simple-number-keys) 
-- [useSimplifiedLogicExpression](/linter/rules/use-simplified-logic-expression) 
-- [useSingleCaseStatement](/linter/rules/use-single-case-statement) 
-- [useSortedClasses](/linter/rules/use-sorted-classes) 
-- [useStrictMode](/linter/rules/use-strict-mode) 
-- [useTopLevelRegex](/linter/rules/use-top-level-regex) 
+- [noAccumulatingSpread](/linter/rules/no-accumulating-spread)
```
drwpow, line 11: Note: there are some trailing whitespace removals fixes that happened automatically. I noticed here and other files, trailing whitespace is inconsistent, i.e. doesn’t seem to be re-added from formatting.

I can revert these “noisy” changes if needed!

drwpow:
> These pages are generated from the code in the main repo. We define rule sources like this: https://github.com/biomejs/biome/blob/80d9b07f2e3bab3b5227a5d21f33fdbb03a1a9ab/crates/biome_js_analyze/src/lint/complexity/no_excessive_nested_test_suites.rs#L61
> 
> We use those rule sources for things like the `biome migrate eslint` command

Oh _ALL_ of it is generated! 😅 Whoops! Will open there then.

drwpow:
Updated PR: https://github.com/biomejs/biome/pull/5872

Issue #2019, opened 2025-04-10
Incorrect bunx and pnpm biome init scripts


Psionyc:
### Documentation URL

https://biomejs.dev/guides/getting-started/

### Description

Seems bunx points to an incorrect package which doesn't seem to work and throws an error

![Image](https://github.com/user-attachments/assets/d2937680-2c34-4c2e-bf05-720f72f50950)

### Expectations

I was expecting a 

```sh
bunx  @biomejs/biome init
pnpx @biomejs/biome init
```

### Code of Conduct

- [x] I agree to follow Biome's Code of Conduct

mantissa7:
I tried to address this but it was closed as working as intended, so this should be closed as wontfix.

PR #2185, opened 2025-05-06
Update PackageManagerBiomeCommand.astro

mantissa7:
Fix docs pointing to biome init instead of @biomejs/biome

## Summary

fixes https://github.com/biomejs/website/issues/2019

### DIFF
--- a/src/components/PackageManagerBiomeCommand.astro
+++ b/src/components/PackageManagerBiomeCommand.astro
@@ -15,13 +15,13 @@ const biomeBin = "biome";
 		<Code code={`npx ${biomePackage} ${command}`} lang="bash"  />
 	</TabItem>
 	<TabItem label="yarn" icon="seti:yarn">
-		<Code code={`yarn ${biomeBin} ${command}`} lang="bash"  />
+		<Code code={`yarn ${biomePackage} ${command}`} lang="bash"  />
 	</TabItem>
 	<TabItem label="pnpm" icon="pnpm">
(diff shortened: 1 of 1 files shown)


dyc3:
I don't think this is right. Both `bunx biome` and `bunx @biomejs/biome` work for me. I also tried `pnpm biome` and `pnpm @biomejs/biome`, and the only one that **didn't** work was `pnpm @biomejs/biome`.

mantissa7:
Oh right. I just tried `bunx biome init` and I got the "package name must be more than 1 character" error
Running `bunx @biomejs/biome init` worked as expected

dyc3:
If you haven't added `@biome/biome` to your `package.json`, then referring to the binary won't work. The instructions on the page tell you to install the package before using it. 
image

mantissa7:
I think this is a confusing paradigm. The purpose of `npx`, and by extension `bunx` is `... to run an arbitrary command from an npm package (either one installed locally, or fetched remotely), in a similar context as running it via npm run.`.

I would expect that an npx command that doesn't find the local command to fetch it and run it from the remote package, since that is the learned behaviour from other packages.

I think the main misunderstanding comes from the different namespaces: `biome` and `@biomejs/biome`. Why does one correctly pull from the remote and the other doesn't? Why are there 2 different namespaces in the first place?

I would expect a global install to have a binary named `biome`, but regardless of whether I was running the JS package locally or remotely I would expect it be the same package/command.

However the main point of this change is the inconsistent commands across the package managers.

PR #2219, opened 2025-05-10
Scanner docs

arendjr:
## Summary

Explain behaviour of the scanner under the `files.includes` docs.

### DIFF
--- a/src/content/docs/reference/configuration.mdx
+++ b/src/content/docs/reference/configuration.mdx
@@ -94,6 +94,28 @@ This means that:
 * `src/special.test.js` **is* processed.
 * `test/special.test.js` **is not* processed.
 
+#### Note on Biome's scanner
+
+Biome has a scanner that is responsible for discovering `.gitignore` files as
+well as indexing projects if any of the rules from the project domain are
+enabled.
+
+The scanner respects `files.includes`, with two exceptions:
+
+* If the project domain or one of its rules is enabled, dependencies in
+  `node_modules` are still read, even if they're ignored in `files.includes`.
+  This is because they may be a valuable source of type information.
+* The scanner will not ignore individual files if they are in a non-ignored
+  directory. This is because files that are co-located with regular
+  (non-ignored) files are likely to be a valuable source of type information,
+  and they are also more likely to form import cycles with other files.
+
+This means that if you use `files.includes` to ignore folders with generated
+code, type information from such folders is inaccessible to the scanner. If you
+do wish to have such type information available, we advise to _not_ ignore these
+folders through `files.includes` and instead ignore them using tool-specific 
+settings such as `linter.includes` and `formatter.includes` only.
+
 ### `files.ignoreUnknown`
 
 If `true`, Biome won't emit diagnostics if it encounters files that it can't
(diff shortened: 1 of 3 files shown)


minht11:
Should we also add note below [project domain](https://next.biomejs.dev/linter/domains/#project) docs about generated files linking here? Current explanation is good, but quite far down inside documentation. Users enabling project domain, likely would not expect types resolving to be impacted by ignoring files.

src/content/docs/reference/configuration.mdx
```diff
@@ -94,6 +94,28 @@ This means that:
...
+
+Biome has a scanner that is responsible for discovering `.gitignore` files as
+well as indexing projects if any of the rules from the project domain are
+enabled.
+
+The scanner respects `files.includes`, with two exceptions:
+
+* If the project domain or one of its rules is enabled, dependencies in
+  `node_modules` are still read, even if they're ignored in `files.includes`.
+  This is because they may be a valuable source of type information.
```
Conaclos, line 107: I wonder if this exception is worth because `node_modules` is already ignored in traversal,
Thus, removing this exception could allow user to have a way of make the scanner ignores `node_modules` if they want.

ematipico: I think it will become part of the deal, and users will have to learn to use it and understand the implications. 

Practically, if a user decides to use `noFloatingPromises`, and they use an async function imported from a library, they expect that rule to work for that function, otherwise it's pretty useless.

It's like TypeScript. You usually ignore `node_modules`, but TSC still looks at it to resolve type declarations

Conaclos: > But remember, contrary to TypeScript, our scanner checks the dependencies only if there are some lint rules that require it (which are all opt-in).

Sure, however we can expect that more and more rules will rely on type info. This will become hard to avoid type info in a such world.

ematipico: We won't know for sure until we try and get the user's feedback in the long run.

Banging our heads trying to find the right implementation is pointless now: it's a new feature, and won't know how it pans out until we try out there in the minefield.

What I'm trying to say is that this might (not) be the right solution for, or maybe it won't/will be your solution, or maybe the right implementation is out there waiting to be discovered by someone else. I believe we spent **enough** energies with the beta, and honestly I'm very satisfied with the current implementation. IMO we should be proud of what we achieved, let's take a break and come back in a few months once we have some feedback at hand.

arendjr: Thanks. I'm sympathetic to both sides of the argument, but I think I agree most with: let's settle on something and see what the feedback is :)

It also occurred to me that we may want to have a configuration akin to [typeRoots](https://www.typescriptlang.org/tsconfig/#typeRoots) at some point (or respect the setting from `tsconfig.json` directly). That may also provide a natural path forward where `node_modules`, which is now a hardcoded exception, will become part of a configurable exception. But that kinda makes me lean into thinking that alignment with our traversal isn't necessarily the best path forward.

Conaclos: Respecting the [`types` setting](https://www.typescriptlang.org/tsconfig/#types) could also improve performance: similarly to TSC we could just scan the concerned packages.

Conaclos: Respecting the [`types` setting](https://www.typescriptlang.org/tsconfig/#types) could also allow us to avoid scanning all `@types/*` packages.

PR #2224, opened 2025-05-11
fix: Added clarification to `biome check --stdin-file-path` CLI docs

magnusriga:
## Summary

Added small entry to `biome check --stdin-file-path` CLI docs.
Makes it clear that `--write` is needed alongside `--stdin-file-path`, for the output to stdout to be the fully fixed file.


ematipico:
@magnusriga you should send the PR against `biomejs/biome`. That part of the file is code generated

dyc3 (approved):

PR #2273, opened 2025-05-17
docs: update override snippet for partial vue/svelte/astro support

dyc3:
## Summary

https://github.com/biomejs/biome/issues/6049 alerted us that this page was outdated. I've fixed the snippet at the end to use the new `includes` and the correct globs. I've also added a couple new rules that I've found need to be disabled via dogfooding biome on my own projects.

### DIFF
--- a/src/content/docs/es/internals/language-support.mdx
+++ b/src/content/docs/es/internals/language-support.mdx
@@ -74,12 +74,16 @@ A partir de la versión `1.6.0`, estos idiomas están soportados **parcialmente*
   {
     "overrides": [
       {
-        "include": ["*.svelte", "*.astro", "*.vue"],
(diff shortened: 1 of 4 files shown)


src/content/docs/es/internals/language-support.mdx
```diff
@@ -74,13 +74,17 @@ A partir de la versión `1.6.0`, estos idiomas están soportados **parcialmente*
...
           "rules": {
             "style": {
               "useConst": "off",
               "useImportType": "off"
-            }
+            },
+            "correctness": {
+  						"noUnusedVariables": "off",
+  						"noUnusedImports": "off"
+  					}
```
siketyan, line 87: Indentation seems broken 👀

siketyan (commented):
Thank you!

siketyan (approved):

PR #2299, opened 2025-05-21
feat: allow to select one rule

ematipico:
## Summary

This PR adds the possibility to select a single rule other than `recommended` and `all`. The rules are code generated, and each time we create a new rule, this file is synchronised.

This should make our life easier, as well as the one of the user.


src/playground/tabs/SettingsTab.tsx
```diff
@@ -961,11 +961,18 @@ function LinterSettings({
...
-						value={lintRules ?? LintRules.Recommended}
-						onChange={(e) => setLintRules(e.target.value as LintRules)}
+						value={lintRules}
+						onChange={(e) =>
+							setLintRules(e.target.value as keyof typeof LINT_RULES)
+						}
 					>
-						<option value={LintRules.Recommended}>Recommended</option>
-						<option value={LintRules.All}>All</option>
+						{Object.values(LINT_RULES).map((value) => {
```
dyc3, line 969: It would be nice if `LINT_RULES` was sorted by name. It would be even better if you could search the list from the selector, but I won't block this PR for that.

dyc3 (changes requested):
I had been meaning to implement something like this. Just hadn't gotten around to it.

src/playground/tabs/SettingsTab.tsx
dyc3: Yes, that was intentional. I'm not blocking on adding search. I am blocking on the sorting. Not having these be sorted is a really bad UX because we have 300+ rules.

ematipico: Ah ok, it wasn't clear

ematipico: Done

dyc3 (approved):

PR #2307, opened 2025-05-24
fix: correct key in overrides in monorepo docs

georgegebbett:
## Summary

The docs suggest using the key `include` in the `overrides` array in the monorepo section. The correct key is `includes` (with a s)


siketyan (approved):
Thank you!

PR #2339, opened 2025-05-29
feat: document monorepo

ematipico:
## Summary

This PR updates the Big Projects page with a small tutorial on how users can set up a monorepo. It also updates the configuration reference.

### DIFF
--- a/src/content/docs/guides/big-projects.mdx
+++ b/src/content/docs/guides/big-projects.mdx
@@ -3,7 +3,7 @@ title: Use Biome in big projects
 description: A small guide how to set Biome in big projects
 ---
 
-import {FileTree} from '@astrojs/starlight/components';
+import {FileTree, Steps} from '@astrojs/starlight/components';
 
 Biome can provide some tools that can help you to use it properly in big projects, such as monorepo or workspaces that contain multiple projects.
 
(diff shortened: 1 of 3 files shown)


src/content/docs/guides/big-projects.mdx
```diff
@@ -95,34 +95,89 @@ Let's jump to `app/backend/biome.json`, where we need to enable the linting:
...
-          }
+    ```
+    This file is called **root configuration** and it sets the base options inside the project. However, nested
+    configurations can decide to adhere to these options or not. Let's see how.
+
+2. Create nested configuration files, one in each package, where is needed. These nested configuration files must have the
+    field `"root"` set to `false`. Failing to do so will result into an error. Also, we want these packages the follow
+    the formatting standards set up in the root configuration. In order to so, we will use a **new microsyntax**
+    available in Biome v2, which is `"extends": "//"`. This syntax tells Biome to extends from the **root configuration**,
+    regardless of where the nested configuration is.
```
dyc3, line 130: nit: "microsyntax" is jargon

```suggestion
    the formatting standards set up in the root configuration. In order to so, we will add `"extends": "//"`, available in Biome v2, to tell Biome to extend the **root configuration**,
    regardless of where the nested configuration is.
```

src/content/docs/guides/big-projects.mdx
```diff
@@ -95,34 +95,89 @@ Let's jump to `app/backend/biome.json`, where we need to enable the linting:
...
+    {
+        "root": false,
+        "extends": "//",
+        "formatter": {
+            "enabled": false
+        }
+    }
+    ```
+    However, when you use the microsyntax `extends: "//"`, you can **omit `"root": true`, because it implies that this configuration isn't a
+    a root configuration, and it wants extends from the root configuration:
```
dyc3, line 160: nit: "microsyntax" is jargon
```suggestion
    However, when you use `extends: "//"`, you can **omit `"root": true`, because it implies that this configuration isn't a
    a root configuration, and it wants extends from the root configuration:
```

dyc3 (commented):
Looks good, just some minor nits.

PR #2342, opened 2025-05-29
feat(api): add versions endpoints

nhedger:
## Summary

This PR introduces two new API endpoints for retrieving both the list of all Biome versions and the latest version. The purpose of these endpoints is to make it easier for first-party tooling to list and find those versions.

Both endpoints support content negotiation to serve either JSON (default) or plain text, depending on whether the `Accept` header is set to `text/plain` or `application/json`.

### Retrieving all versions

The `/api/versions` endpoint returns the list of all Biome versions, sorted in descending order (latest first). By default it only includes stable versions, but pre-release versions can be included with `?prereleases=true`

```sh
# Get all stable versions as JSON
curl https://biomejs.dev/api/versions

# Get all stable versions as plain text
curl -H 'Accept: text/plain' https://biomejs.dev/api/versions

# Get all stable + prereleases as JSON
curl https://biomejs.dev/api/versions?prereleases=true

# Get all stable + prereleases as plain text
curl -H 'Accept: text/plain' https://biomejs.dev/api/versions?prereleases=true
```

### Retrieving the latest version

The `/api/versions/latest` endpoint returns the latest version of Biome. By default, the latest stable version is returned, but the latest prerelease can be returned as well with `?prerelease=true`.

```sh
# Get the latest stable as JSON
curl https://biomejs.dev/api/versions/latest

# Get the latest stable as plain text
curl -H 'Accept: text/plain' https://biomejs.dev/api/versions/latest
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,6 +75,7 @@
   },
   "packageManager": "pnpm@10.10.0",
   "dependencies": {
+    "@biomejs/version-utils": "^0.4.0",
     "starlight-links-validator": "^0.16.0"
   }
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -8,25 +8,28 @@ importers:
 
   .:
     dependencies:
+      '@biomejs/version-utils':
+        specifier: ^0.4.0
+        version: 0.4.0
       starlight-links-validator:
         specifier: ^0.16.0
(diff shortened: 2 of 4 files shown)


ematipico (approved):
Looks good! I honestly don't know if we have a cap on our function invocations, so we should make sure if we have it. And if so, to not exceed it

nhedger:
That came to mind. We could also generate it statically once a day.

PR #2359, opened 2025-06-07
feat: add domains to table of rules

yacinehmito:
## Summary

Adds a column "Domains" to the table of rules.

I introduced this because I was getting annoyed at having to open each rule's page to determine which domain(s) it belonged to, and thus whether it was relevant for my codebase.


ematipico:
Thank you @yacinehmito. You must run `pnpm codegen:rules` so we can see the result.
As a side note, the reason why that table is "slim" is that it gets to navigate and read on mobile. The addition of a new table, which **isn't** there for all rules, might not be a good addition.

yacinehmito:
Sorry, I did not realize that generated code has to be committed. It's fixed.

> As a side note, the reason why that table is "slim" is that it is difficult to navigate and read on mobile. The addition of a new table, which isn't there for all rules, might not be a good addition.

I agree, it's too narrow. I could make the "properties" cell "double-lined", stacking the icons on top of the list of domains. What do you think?

dyc3:
I like that idea. Try it, we'll see how it looks.

yacinehmito:
I tried both. Putting it in _Properties_ was really ugly. The line would break at the worse points and it would feel bloated. Appending to the description on the other hand looks fine, so I kept that. Let me know what you think.

dyc3 (commented):
How it currently looks:
image

nit: I think it would be more glance-able if the "Domain: ..." part was it's own block. Something like this?
image

I still think it would be better in the Properties. Perhaps it would be better if it was in it's own block, and we omit the "Domains:" prefix?

image

PR #2384, opened 2025-06-09
Zed: Switch configuration from global to language specific

notpeter:
## Summary

Biome's previous recommend Zed settings include settings which were not language scoped, as a result, users regularly reported broken formatters for non-biome supported languages (Rust, etc).  It also means that biome-language-server **always** runs which is not a good default.

You'll notice that all the examples in the Zed Docs are scoped to an individual language, ex:
- https://zed.dev/docs/configuring-languages#configuring-formatters (`formatter`)
- https://zed.dev/docs/configuring-languages#choosing-language-servers (`language_servers`)
- https://zed.dev/docs/configuring-languages#setting-up-linters (`code_actions_on_format`)

I understand this is more verbose but it is also more explicit.

Let me know if there's anything I should do to the translated versions of the docs (e.g. remove the out-of-date translations so user's are not see outdataed docs which will cause trouble). 

See also:
- https://github.com/zed-industries/zed/issues/29300

### DIFF
--- a/src/content/docs/reference/zed.mdx
+++ b/src/content/docs/reference/zed.mdx
@@ -35,13 +35,20 @@ To use the language server as a formatter, specify biome as your formatter in th
 ```jsonc
 // settings.json
 {
-  "formatter": {
-    "language_server": {
-      "name": "biome"
-    }
+  "languages": {
+    "JavaScript": { "formatter": { "language_server": { "name": "biome" } } },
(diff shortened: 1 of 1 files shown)


ematipico:
That's unexpected. I was under the impression that the extension could control when to activate based on the language IDs provided here https://github.com/biomejs/biome-zed/blob/main/extension.toml#L28-L40

Isn't that the case?

PR #2442, opened 2025-06-17
feat: version selector for v2 and next, add banner for outdated docs

siketyan:
## Summary

Updated the version selector to have v2 (now biomejs.dev) and next. Also added a banner for navigating to the latest docs.


src/routeData.ts
siketyan, line 1: I am not sure if this is the best way to add banner to every single page on Astro

arendjr (approved):

PR #2515, opened 2025-06-23
docs: add note about external packages in conjunction with biome action

Willem-Jaap:
**Add documentation note for using external packages with Biome CI action**

This improves the CI recipe documentation by adding a note on handling Biome configs that rely on external dependencies/configs. It includes a snippet showing how to set up Node.js and install dependencies before running biome ci.


ematipico (approved):
Thank you

PR #2522, opened 2025-06-24
Update docs: schema & nested config notes

kenrick95:
## Summary

- Updated the example schema to use 2.0.5 instead of the 2.0.0-beta version
- From my understanding, since 2.0.0 Biome has support for nested configuration, so I updated the note to mention that instead of misleading people saying that it's not supported


siketyan (approved):
Nice catch, thank you!

PR #2604, opened 2025-07-01
Extend scanner documentation

arendjr:
## Summary

This reorganises and extends the scanner documentation, including the new "targeted" scanner behaviour.

Also added `nx.json` and `project.json` to the list of known files.


ematipico:
> Not sure why it complains about the relative link? It should exist...

Try `/reference/configuration/#filesexperimentalscannerignores` (with the trailing slash)

PR #2610, opened 2025-07-03
docs: fix manual installation instructions

ridwanyinus:
The previous installation instructions used outdated GitHub release tag formats (cli/v <version>), which are no longer supported in Biome v2+. This caused curl and PowerShell to download broken 9-byte files instead of the actual CLI binaries.

The motivation for this change is to ensure users can install Biome successfully on macOS, Linux, and Windows without needing to dig through GitHub Releases or Discord for the correct links. This improves onboarding and prevents unnecessary confusion.

This PR updates the URLs to use the new @biomejs/biome@<version> scoped release tag format, which aligns with the current structure of Biome v2+ on GitHub.

### DIFF
--- a/src/content/docs/es/guides/manual-installation.mdx
+++ b/src/content/docs/es/guides/manual-installation.mdx
@@ -65,15 +65,15 @@ Para instalar Biome, descarga el ejecutable para tu plataforma de la [última ve
 
 ```shell
 # macOS arm (M1 o posterior)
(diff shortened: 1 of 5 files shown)


nhedger:
Thanks !

dyc3 (commented):
Good catch! Would you mind updating the links on the pages for other languages as well?

ridwanyinus:
> Good catch! Would you mind updating the links on the pages for other languages as well?

I'll gladly do that sire.

PR #2664, opened 2025-07-08
fix(playground): unexpected "Not supported" output

siketyan:
## Summary

`FileFeaturesResult.featuresSupported` is defined as `{ [K in FeatureKind]?: SupportKind }` but actually it's a `Map` object. Reverted some changes and added `ts-expect-error` for now. I'll fix the type definition later to match the actual value.


siketyan:
I have already merged the fix for the type mismatch, closing this workaround.

PR #2786, opened 2025-07-27
chore: fix typos

noritaka1166:
## Summary
fix typos


src/pages/metadata/rules.json.js
siketyan, line 1: This file is auto-generated from the source code and not intended to be edited by hand. Can you send a PR to the biomejs/biome repository instead?

siketyan (changes requested):

src/pages/metadata/rules.json.js
noritaka1166: OK!  
I reverted it.

siketyan (approved):
Thank you!

PR #2816, opened 2025-08-01
docs: tweak the text that warns the user a rule is in nursery

dyc3:
## Summary

A small change to make it more immediately clear what nursery means.


ematipico:
I would also add that these rules don't follow semver until they are outside of that group

PR #2823, opened 2025-08-03
fix: Correct nursery links in rules

Jiralite:
## Summary

Every nursery rule that says it is part of the group contains a dead link. Fixed the code generation to point to the correct link.


siketyan (approved):
Thank you!

Issue #2820, opened 2025-08-02
Add Russian translation


matvasenin:
Hi everyone!
I'm excited about Biome and would like to contribute by adding Russian localization support.
Thanks for your work!

PR #2824, opened 2025-08-03
docs(ru): add index.mdx

matvasenin:
### 🛠️ Changes
- Add `src/content/docs/ru/index.mdx`
### 💾 Affected Files
- `src/content/docs/ru/index.mdx`
### ❓ Questions
- Does `public/_redirects` need to be updated in order to redirect 404s to a localized page?
- Is it intended that components such as `SiteTitle.astro` cannot be localized, or can I implement this?
### 📋 Related issue: #2820

### DIFF
--- a/src/content/docs/ru/index.mdx
+++ b/src/content/docs/ru/index.mdx
@@ -1,0 +1,232 @@
+---
+title: Biome
+head:
+  - tag: title
+    content: Biome, веб-инструментарий
+template: splash
+description: Форматирование, линтинг и многое другое за долю секунды.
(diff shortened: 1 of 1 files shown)


dyc3 (approved):

ematipico:
> Does public/_redirects need to be updated in order to redirect 404s to a localized page?

Yes

>  Is it intended that components such as SiteTitle.astro cannot be localized, or can I implement this?

We haven't had the chance to make it localized friendly, so if you want to improve it, you're more than welcome

PR #2857, opened 2025-08-06
docs: don't use `/**` to ignore folders

arendjr:
## Summary

This now works "properly" in 2.2.

### DIFF
--- a/src/content/docs/guides/configure-biome.mdx
+++ b/src/content/docs/guides/configure-biome.mdx
@@ -213,7 +213,7 @@ In the following example, we tell Biome to include files, excluded the `dist` di
   "files": {
     "includes": [
       "**",
-      "!**/dist/**",
+      "!**/dist",
       "!**/*.generated.js"
     ]
   }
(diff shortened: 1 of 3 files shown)


src/content/docs/reference/configuration.mdx
```diff
@@ -1325,10 +1325,18 @@ following syntax in globs:
...
+  you need to use the `/**` suffix to match those files. But if you want to
+  _ignore_ all files in a folder, you may do so without the `/**` suffix. We
+  recommend ignoring folders without the trailing `/**`, to avoid needlessly
+  traversing it, as well as to avoid the risk of Biome loading a `biome.json` or
+  a `.gitignore` file from an ignored folder.
 
 Some examples:
 
 - `dist/**` matches the `dist/` folder and all files inside it.
+- `!dist` ignores the `dist/` folder and all files inside it.
```
ematipico, line 1339: Should we document what `dist` now means, too? It has different semantics... and it's probably useless, right? I suppose that adding `src` inside `file.includes` won't match any file

src/content/docs/guides/configure-biome.mdx
```diff
@@ -213,7 +213,7 @@ In the following example, we tell Biome to include files, excluded the `dist` di
   "files": {
     "includes": [
       "**",
-      "!**/dist/**",
```
ematipico, line 216: I realised that what we say at line  `209` isn't entirely correct. `!**/dist` esclused **any** `dist` directory. It might be a nitpick, should we update it? Or maybe we should update the glob to `!dist`

src/content/docs/reference/configuration.mdx
arendjr: I'm not entirely sure what you're getting at... isn't that what I documented in the paragraph above?

PR #2868, opened 2025-08-07
feat(ja): update menu translations

siketyan:
## Summary

Updated translations of menu items.
I also fixed the "Docs" link in the navbar not navigating to the localised page.


unvalley (approved):

unvalley:
@siketyan The @textlint/text plugin makes it possible, but prh.yml is also applied to TS files, e.g., astro.config.ts (checking things like changing “formatter” to “フォーマッタ” in English menus…).
So, I don’t think we need to do that. It doesn’t help us much.

Issue #2877, opened 2025-08-09
📚 Title JA blog dates show HTML-escaped slashes (e.g., 2025&#x2F;07&#x2F;08)


ryu1013-job:
## Documentation URL

https://biomejs.dev/ja/blog/

### Overview
When the website's language is set to Japanese, the forward slash `/` in dates (e.g., on blog posts) is incorrectly HTML-escaped as `&#x2F;` and displays incorrectly.

### Steps to Reproduce
1. Go to a Biome page with a date, such as a blog post (e.g., https://biomejs.dev/blog/biome-v2-1/)
2. Use the language switcher in the top-right corner to change the language to Japanese (日本語).
3. Observe the date displayed below the post title.

**Screenshot**
<img width="1552" height="987" alt="Image" src="https://github.com/user-attachments/assets/913cc234-668a-4586-aad8-5406cbfcbe76" />

### Environment
- OS: macOS
- Browser: Chrome, Safari

### Expected Behavior
The date should be displayed correctly with forward slashes, for example: `2025/07/08` or a locale-appropriate format.

### Actual Behavior
The date is displayed with the forward slash escaped as an HTML entity, for example: `2025&#x2F;07&#x2F;08`.

## Code of Conduct

- [x] I agree to follow Biome's Code of Conduct

PR #2878, opened 2025-08-09
fix(ja): use long date style for ja to avoid “/” in blog dates

ryu1013-job:
## Summary
**Fix JA blog dates showing HTML-escaped slashes.**

### What
- Override JA i18n for blog date
  - Update `ja.json`

### Why
- Default JA format uses “/”, which is rendered as &#x2F; and appears as mojibake.

### Scope/Impact
- Affects JA blog index and post pages only.
- Other locales remain unchanged.

## Screenshots
| Before | After |
|--------|--------|
| Screenshot 2025-08-10 at 2 31 48 | Screenshot 2025-08-10 at 2 44 26 | 

Fixes https://github.com/biomejs/website/issues/2877

### DIFF
--- a/src/content/i18n/ja.json
+++ b/src/content/i18n/ja.json
@@ -1,5 +1,6 @@
 {
   "title.docs": "ドキュメント",
   "title.enterprise": "エンタープライズ",
-  "title.playground": "プレイグラウンド"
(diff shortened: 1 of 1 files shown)


siketyan (approved):
Nice catch, thank you!

PR #2891, opened 2025-08-11
Docs lint

polyzen:
## Summary

- Rename `create-a-extension` pages to `create-an-extension`
- Remove superfluous file


ematipico (changes requested):
You must update the sidebar too. It's now giving a 404

ematipico:
I'm not talking about the environment variable page, I'm talking about the page you renamed. Check the `astro.config.mjs`, it still points to the old page

ematipico:
I'm not talking about the environment variable page, I'm talking about the page you renamed. Check the `astro.config.mjs`, it still points to the old page

siketyan:
Also we should configure redirects so the existing links won't break

polyzen:
> Also we should configure redirects so the existing links won't break

Using https://github.com/biomejs/website/blob/main/public/_redirects?

polyzen:
> I'm not talking about the environment variable page, I'm talking about the page you renamed. Check the `astro.config.mjs`, it still points to the old page

Added in the force-push: https://github.com/biomejs/website/compare/abaa579be5823814c1dcf06e62dc116c81a1ee81..0501854c197b06e6e1776cbc689fa3dc6450deb3

https://github.com/biomejs/website/pull/2891/commits/0501854c197b06e6e1776cbc689fa3dc6450deb3 removes `src/content/docs/reference/environment_variables.md` as it's a duplicate of `environment-variables.md`. `env-variables.rs` creates `environment_variables.md`, so I'm wondering what should be done here.

ematipico:
> > Also we should configure redirects so the existing links won't break
> 
> Using https://github.com/biomejs/website/blob/main/public/_redirects?

We prefer using Astro for redirects

https://github.com/biomejs/website/blob/main/astro.config.ts#L96-L97

PR #2940, opened 2025-08-15
Fix example to integrate biomejs with gitlab-ci

Pascalmh:
## Summary

The example of how to integrate biome with gitlab-ci was outdated, this fixes the example code

- the **code-quality.json** needs to be passed as `artifacts.reports.codequality`
- the note about the `src` directory seems to be outdated as there is no reference to a `src` directory in the example-code

### DIFF
--- a/src/content/docs/recipes/continuous-integration.mdx
+++ b/src/content/docs/recipes/continuous-integration.mdx
@@ -86,8 +86,9 @@ lint:
         - biome ci --reporter=gitlab --colors=off > /tmp/code-quality.json 
         - cp /tmp/code-quality.json code-quality.json
     artifacts:
-        paths:
-            - code-quality.json    # Collect the code quality report as an artifact
+      reports:
+        codequality:
+          - code-quality.json    # Collect the code quality report artifact
     rules:
         - if: $CI_COMMIT_BRANCH    # Run job for commits on branches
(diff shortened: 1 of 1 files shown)


src/content/docs/recipes/continuous-integration.mdx
```diff
@@ -86,13 +86,10 @@ lint:
         - biome ci --reporter=gitlab --colors=off > /tmp/code-quality.json 
         - cp /tmp/code-quality.json code-quality.json
```
Pascalmh, line 87: Question: when the `biome ci --reporter=gitlab --colors=off > /tmp/code-quality.json` line failes, it seems like gitlab-ci does not continue to the next line, so `cp /tmp/code-quality.json code-quality.json` is never run.

Why is the report first put into the `/tmp/`-dir only to be copied back into the working directory?

Why not directly this?

```bash
biome ci --reporter=gitlab --colors=off > code-quality.json
```

dyc3: It's necessary because the pipe creates the file, and biome can process JSON files. Biome will pick up the file, and emit parsing diagnostics for it.

image

dyc3:
> Why recommend to use `latest` tag here

Because we haven't set up any codegen to keep the latest version up to date.

ematipico (approved):

PR #2962, opened 2025-08-20
docs: add `css.parser.tailwindDirectives`

dyc3:
## Summary

docs PR for https://github.com/biomejs/biome/pull/7164


Jayllyz:
It might be useful to specify whether this applies only to Tailwind v4 syntax or to all syntax versions, given that `v4` is not very old

PR #2965, opened 2025-08-21
feat: rule suggestion

Netail:
## Summary

Added a note on te bottom of the [language] rules about contributing or suggesting creating a rule suggestion discussion

### DIFF
--- a/codegen/src/shared.rs
+++ b/codegen/src/shared.rs
@@ -19,7 +19,7 @@ pub fn add_codegen_rule_suggestion(content: &mut Vec<u8>) -> anyhow::Result<()> 
     writeln!(content)?;
     writeln!(
         content,
-        "Missing a rule? Help us by [contributing](https://github.com/biomejs/biome/blob/main/CONTRIBUTING.md) or create a rule suggestion [here](https://github.com/biomejs/biome/discussions/categories/rule-suggestion)."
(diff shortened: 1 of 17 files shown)


codegen/src/shared.rs
```diff
@@ -14,3 +14,12 @@ pub fn add_codegen_disclaimer_frontmatter(content: &mut Vec<u8>) -> anyhow::Resu
     writeln!(content)?;
     Ok(())
 }
+
+pub fn add_codegen_rule_suggestion(content: &mut Vec<u8>) -> anyhow::Result<()> {
+    writeln!(content)?;
+    writeln!(
+        content,
+        "Missing a rule? Help us by [contributing](https://github.com/biomejs/biome/blob/main/CONTRIBUTING.md) or create a rule suggestion [here](https://github.com/biomejs/biome/discussions/categories/rule-suggestion)."
```
dyc3, line 22: This should probably also link to the `biome_analyze` contributing.md as well.

Netail: The root contributing.md links to the biome_analyze under "Analyzers and lint rules"

ematipico: > The root contributing.md links to the biome_analyze contributing.md under "Analyzers and lint rules" tho 🤔

True, however I found out that sometimes people don't reach that part. By default users are lazy, so we should match the lazyness.

ematipico (approved):

ematipico:
Thank you @Netail

PR #2967, opened 2025-08-21
docs: update `files.includes` doc

baptiste-prunot:
## Summary

Add a caution block regarding the usage of `!` to exclude directories in `files.includes`. The intent is to emphasize that using `!test` will work in `files.includes` and will exclude completely the `test` directory, but will not have that effect in other places where `includes` exists (like `linter`, `formatter`, etc.)

I'm suggesting this update because of https://github.com/biomejs/biome/issues/7262

### DIFF
--- a/src/content/docs/reference/configuration.mdx
+++ b/src/content/docs/reference/configuration.mdx
@@ -102,6 +102,10 @@ This example specifies that:
 4. ... _except_ when it occurs in the folder named `test`, because _no_ files
    inside that folder are processed.
 
+:::caution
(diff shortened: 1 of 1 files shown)


src/content/docs/reference/configuration.mdx
```diff
@@ -102,6 +102,10 @@ This example specifies that:
 4. ... _except_ when it occurs in the folder named `test`, because _no_ files
    inside that folder are processed.
 
+:::caution
```
ematipico, line 105:
```suggestion
:::note
```

ematipico (approved):
Just a minor tweak. Thank you!

PR #3007, opened 2025-08-27
docs: update `html.formatter.indentScriptAndStyle`

dyc3:
## Summary

Docs update for https://github.com/biomejs/biome/pull/7333


src/content/docs/reference/configuration.mdx
```diff
@@ -1378,9 +1378,11 @@ Whether to account for whitespace sensitivity when formatting HTML (and its supe
 
 ### `html.formatter.indentScriptAndStyle`
 
-Whether to indent the `<script>` and `<style>` tags for HTML (and its super languages).
+*Since 2.3*
 
-> Default: `true`
+Whether to indent the content of `<script>` and `<style>` tags for HTML-ish templating languages (Vue, Svelte, etc.).  When true, the content of `<script>` and `<style>` tags will be indented one level.
```
ematipico, line 1383: A couple of things:
- we should list **exactly** the languages/files. The "etc." is too open ended and could cause false expectations
- "one level" may not be clear enough. In HTML those tags might be on different levels of indentation. Does that mean the content is still at level "one" of indentation?

ematipico (approved):

PR #3014, opened 2025-08-28
docs(pt-BR): update configure-biome guide

dieegofrr:
## Description
Updates the Portuguese (Brazil) translation of the configure-biome guide.

## Changes
- Updated content to match latest English version
- Fixed outdated information

### DIFF
--- a/astro.config.ts
+++ b/astro.config.ts
@@ -192,7 +192,7 @@ export default defineConfig({
 								fr: "Configurer Biome",
 								ja: "Biome の設定",
 								"zh-CN": "配置 Biome",
-								"pt-BR": "Configurar Bioma",
+								"pt-BR": "Configurar o Biome",
 								uk: "Налаштування Biome",
 								ru: "Настройка Biome",
 							},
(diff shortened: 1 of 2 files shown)


astro.config.ts
```diff
@@ -192,7 +192,7 @@ export default defineConfig({
 								fr: "Configurer Biome",
 								ja: "Biome の設定",
 								"zh-CN": "配置 Biome",
-								"pt-BR": "Configurar Bioma",
+								"pt-BR": "Configurar o Biome",
```
dyc3, line 195: This isn't blocking, I'm just curious. What's the difference for this change? Machine translation translates it as exactly the same thing. Is it like "Configuring Biome" vs "Configuration of Biome"?

dyc3 (approved):
Thank you!

astro.config.ts
dieegofrr: Hi! Thanks for the question.

The change corrects a couple of points. In Portuguese, the project name is a proper noun, **Biome**, and the previous translation used the word "**bioma**", which is a different word that means an ecological biome (like a rainforest).

Also, to correctly refer to the project's name, we naturally add the article "**o**" ("the") before it. So, "**Configurar o Biome**" is the grammatically correct and most common way to say "Configure Biome" in Portuguese.

PR #3030, opened 2025-09-03
docs: remove duplicate `html.formatter.enabled`

mcecode:
## Summary

I was just checking out the docs and noticed the duplicate. I removed the one saying that the default is `true` since the HTML formatter is still experimental and whose description did not match the format of the other `*.formatter.enabled` entries.

Feel free to close this if I missed something and the duplication is intentional.


dyc3:
Ah, hang on I didn't catch that it was a duplicate. Github wasn't showing me that in the diff

PR #3052, opened 2025-09-06
feat: multi file lint rule codegen support

ryan-m-walker:
## Summary

This PR modifies the website codgen process to utilize the `file=<path>` code block annotation for setting rendered block filenames as suggested in [this comment](https://github.com/biomejs/biome/pull/7399#discussion_r2323990001) and also adds support for rendering multi-file lint rule diagnostics. Much of the functionality is similar to [this PR](https://github.com/biomejs/biome/pull/7399) but ported to this repo.

The approach here is a bit different from the one in the main Biome repo, opting instead run a second parse pass to collect all files for the file system if we detect the need for it. This seemed like the better approach here since with the codegen process we are dealing with writers that write while we parse making the deferred evaluation approach used in the other PR much less practical or performant. File system parsing is run lazily though only when we detect the need for it so should hopefully add very little extra overhead. I'm open to exploring other approaches if there are any suggestions though.

@arendjr I mostly just copied over the `get_test_services` function over from the other repo with your changes as well. Maybe it would make sense to add this to the `biome_test_utils` package in the main repo?  If it would be helpful I could do so in a follow up PR if you would like.

**Example:**
   
Screenshot 2025-09-05 at 11 09 46 PM


codegen/src/lintdoc.rs
```diff
@@ -1608,16 +1647,24 @@ fn write_action<L: ServiceLanguage>(
 /// Parse and analyze the provided code block, and asserts that it emits
 /// exactly zero or one diagnostic depending on the value of `expect_diagnostic`.
 /// That diagnostic is then emitted as text into the `content` buffer
+#[allow(clippy::too_many_arguments)]
```
ryan-m-walker, line 1650: hopefully this is acceptable, i saw it used a few other places in this file already 😅

ematipico: Yeah it's fine, we don't have a strong linting policy here :)

codegen/src/lintdoc.rs
```diff
@@ -1515,9 +1544,19 @@ impl FromStr for CodeBlockTest {
...
+            file_path: None,
         };
 
         for token in tokens {
+            if let Some(file) = token.strip_prefix("file=") {
+                if file.is_empty() {
+                    bail!("The 'file' attribute must be followed by a file path");
+                }
+
+                test.file_path = Some(file.to_string());
```
ematipico, line 1556: This assumes that the code examples have only one file. If the code examples use multiple files, is that an error? 

If that's an error, do we catch that in the main repository?

If it's not an error, we should this be a Vector?

codegen/src/lintdoc.rs
```diff
@@ -1608,16 +1647,24 @@ fn write_action<L: ServiceLanguage>(
...
     buffer: &mut HTML<&mut Vec<u8>>,
+    file_system: &HashMap<String, String>,
     to_print_kind: ToPrintKind,
 ) -> Result<()> {
-    let file_path = format!("code-block.{}", test.tag);
+    let file_path = {
+        if let Some(file_path) = &test.file_path {
+            normalize_file_path(file_path)
+        } else {
+            format!("code-block.{}", test.tag)
```
ematipico, line 1665: We don't validate if the file has or doesn't have an extension. Do we check for that in the main repository ?

codegen/src/lintdoc.rs
```diff
@@ -1979,3 +2025,131 @@ fn extract_summary_from_rule(content: &str) -> String {
 
     events_to_text(events)
 }
+
+/// Parses markdown documentation to extract in-memory file systems scoped by section.
+/// To match the way we do rule checking in the Biome repo, we scope each
+/// file system to a markdown section. This allows lint rules to access
+/// multiple related files within the same documentation section while
+/// keeping different examples isolated.
+fn parse_file_system(docs: &'static str) -> Result<HashMap<usize, HashMap<String, String>>> {
```
ematipico, line 2034: It's not very clear what a "section" is. You might want to clarify that.

Also, it would be great if we explain the business logic of the parsing somewhere

src/content/docs/linter/rules/no-floating-promises.mdx
```diff
@@ -59,16 +59,16 @@ one of the following ways:
...
 
-```file=async-fn2.ts
+```ts title='async-fn2.ts'
```
ematipico, line 71: This is a great improvement!!

ematipico (commented):
I left a few questions, but overall it's looking amazing!

codegen/src/lintdoc.rs
arendjr: I think this logic gets invoked _per code block_, no? A single code block always represents a single file, but there can be multiple code blocks in an example.

arendjr (approved):
Great work!

Indeed it would be nice if you could expose `get_test_services()` from `biome_test_utils` so we don't need to duplicate it here, but it's fine if you do that in a separate PR :+1:

codegen/src/lintdoc.rs
ryan-m-walker: yeah each one of these tests is for a single code block, initialize [here](https://github.com/biomejs/website/blob/9b63a722214d46e2e4e200e234476f2809e7ce4d/codegen/src/lintdoc.rs#L1223) in the code start event and evaluated [here](https://github.com/biomejs/website/blob/9b63a722214d46e2e4e200e234476f2809e7ce4d/codegen/src/lintdoc.rs#L1246) in the end code block event

codegen/src/lintdoc.rs
ryan-m-walker: I tried to improve the documentation to be more clear. Let me know if you think it's better or could be improved more

ematipico:
After reading the business logic and looking at the final result:l https://deploy-preview-3052--biomejs.netlify.app/linter/rules/no-import-cycles/#invalid , I have the following question: **do we need headings for each file?** They seem redundant if you look at the web page, because the file name are already part of code block

ryan-m-walker:
@ematipico I opened a small clean up [PR here](https://github.com/biomejs/biome/pull/7459) that would remove the filenames in favor of the `file=<path>` attributes instead and remove the redundant header in the UI. Those headers shouldn't be needed and are just left over from before the file path stuff was set up. Does that cover your concern?

ematipico (approved):
Looks good! Happy to merge after the conflicts have been resolved

ryan-m-walker:
@ematipico merge conflicts resolved!

PR #3125, opened 2025-09-19
Fix: Remove duplicated organizeImports configuration example

leaysgur:
## Summary

https://biomejs.dev/assist/actions/organize-imports/#import-and-export-groups

The same content seemed to be repeated.

image

### DIFF
--- a/src/content/docs/assist/actions/organize-imports.mdx
+++ b/src/content/docs/assist/actions/organize-imports.mdx
@@ -274,23 +274,6 @@ Given the following configuration...
 
 ```json title='biome.json'
 {
-    "assist": {
-        "actions": {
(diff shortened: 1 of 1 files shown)


ematipico:
This file is generated from the source code of the rule. There's a comment at the top of the file that says that. I'll close the PR

leaysgur:
Oh, I'm sorry, I didn't notice that. Thank you for your triage.

FYI: There were no duplicates in the source code.

https://github.com/biomejs/biome/blob/3f06e19c6eb8476ad9de4e3dac00c50a2d6f0aed/crates/biome_js_analyze/src/assist/source/organize_imports.rs#L273-L291

PR #3370, opened 2025-10-31
fix: align schema generation

Netail:
## Summary

Align schema generation with https://github.com/biomejs/biome/blob/main/xtask/codegen/src/generate_schema.rs

Related # https://github.com/biomejs/biome/discussions/7865


Netail:
https://github.com/biomejs/website/pull/3378

Issue #3640, opened 2025-12-14
📚 Non-recommended rules should be explicitly labeled as such


PaulRBerg:
### Documentation URL

https://biomejs.dev/linter/rules/use-exhaustive-dependencies/

### Description

It would be helpful if non-recommended rules were explicitly labeled as such on the Biome website.

### Expectations

When I go look at a rule like [`useExhaustiveDependencies`](https://biomejs.dev/linter/rules/use-exhaustive-dependencies/), it should be immediately obvious that the rule is **not recommended**:

<img width="300" src="https://github.com/user-attachments/assets/6c863b62-5df9-4cc2-a3cb-f4d9b0396ab7" />

Just like when I look at a rule like [`noUnusedVariables`](https://biomejs.dev/linter/rules/no-unused-variables/), it is indeed obvious that it is a recommended rule:

<img width="300" src="https://github.com/user-attachments/assets/64012f2b-5e41-441d-aee6-e09e806e87d8" />

Explicit docs > implicit docs

Note: issue https://github.com/biomejs/website/issues/2279 is related, but I was confusing two websites there. This issue is clear and targeted on a particular problem I have continued to encounter while browsing the Biome website.

### Code of Conduct

- [x] I agree to follow Biome's Code of Conduct

PR #3641, opened 2025-12-14
feat(docs): add explicit labels for non-recommended rules

PaulRBerg:
Non-recommended linter rules now display an information icon in rules tables and explicit text on individual rule pages. This makes it clearer which rules are disabled by default without being recommended.

Nursery and domain-scoped rules are excluded from this indicator as they have their own distinct labeling.

Closes https://github.com/biomejs/website/issues/3640


PaulRBerg:
Addressed your feedback @ematipico in the latest git commit

PR #3677, opened 2025-12-16
add missing level to noExcessiveCognitiveComplexity

bodadotsh:
## Summary

Without the `level` property, the `maxAllowedComplexity` have no effect on `noExcessiveCognitiveComplexity`


dyc3:
Per the note at the top of the file:

> Don't modify this file manually. This file is auto generated from source, and you will lose your changes next time the website is built.
> Head to the `biomejs/biome` repository, and modify the source code in there.

PR #3774, opened 2026-01-02
docs: Clarify Biome's ignore comment functionality

vitorbonamigosassi:
Updated documentation to clarify Biome's support for ignoring files to keep up to date with v2.

## Summary

The Formatter's introduction in the documentation states Biome does not provide a comment to ignore the whole file. A behavior which was implemented on [Biome v2](https://next.biomejs.dev/blog/biome-v2/#improved-suppressions).


dyc3 (approved):
Good catch, thanks!

ematipico (changes requested):
That's not true, the formatter doesn't support the ignore all yet.

Try a playground, you'll see

PR #4093, opened 2026-03-23
docs: add SVG to language support page

hammadxcm:
> This PR was written with AI assistance (Claude Code).

## Summary

Add SVG to the language support documentation page.

## Description

- Added SVG as a row in the supported languages table with full support for parsing, formatting, and linting.
- Added a short "SVG support" section explaining that Biome treats `.svg` files as HTML.

## Related

Related to biomejs/biome#9604 and biomejs/biome#9608.


siketyan:
We don't support SVG yet. We have a few rules that applies to SVG, but it applies only to `<svg>` elements **inside HTML files**. It doesn't mean we support standalone SVG files.

siketyan:
My bad, I didn't notice the related PR https://github.com/biomejs/biome/pull/9608. I'll keep this pr open too.

PR #4116, opened 2026-04-02
feat(playground): add JSON editor for Playground settings

mkosei:
## Summary
Closes #4113 
- add an `Edit as JSON` entry point in the Playground settings tab
- add a native `dialog`-based JSON editor for editing Playground settings
- expose settings as a `biome.json`-like configuration that can also be copied

## Notes

- the JSON includes a placeholder `$schema` value
- users can update the `$schema` version manually if needed
- this keeps the existing worker behavior unchanged

### DIFF
--- a/src/playground/buildBiomeConfiguration.ts
+++ b/src/playground/buildBiomeConfiguration.ts
@@ -25,9 +25,15 @@ function getBiomeIndentStyle(indentStyle: PlaygroundSettings["indentStyle"]) {
 }
 
 function getPlaygroundIndentStyle(indentStyle: IndentStyle | undefined) {
-	return indentStyle === IndentStyle.Space
-		? IndentStyle.Space
-		: defaultPlaygroundState.settings.indentStyle;
+	if (indentStyle === IndentStyle.Space) {
+		return IndentStyle.Space;
+	}
+
+	if (indentStyle === IndentStyle.Tab) {
+		return IndentStyle.Tab;
+	}
+
+	return undefined;
 }
 
 function getBiomeAttributePosition(
@@ -41,9 +47,15 @@ function getBiomeAttributePosition(
 function getPlaygroundAttributePosition(
 	attributePosition: AttributePosition | undefined,
 ) {
-	return attributePosition === AttributePosition.Multiline
-		? AttributePosition.Multiline
-		: defaultPlaygroundState.settings.attributePosition;
+	if (attributePosition === AttributePosition.Multiline) {
+		return AttributePosition.Multiline;
+	}
+
+	if (attributePosition === AttributePosition.Auto) {
+		return AttributePosition.Auto;
+	}
+
+	return undefined;
 }
 
 function getBiomeExpand(expand: PlaygroundSettings["expand"]) {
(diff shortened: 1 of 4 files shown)


ematipico (commented):
It would be nice to make the modal full screen on mobile

Image

Also, after the refactor, we are using so many repeated strings. We should re-use the ones we already have

src/playground/components/SettingsJsonEditorModal.tsx
```diff
@@ -0,0 +1,176 @@
...
+	function applyJsonSettings() {
+		try {
+			const parsed = JSON.parse(configurationAsJson) as Configuration;
+			onApply(parseBiomeConfiguration(parsed, settings));
+			onClose();
+		} catch (error) {
+			setJsonError(
+				error instanceof Error ? error.message : "Invalid JSON settings.",
+			);
+		}
```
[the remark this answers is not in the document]
mkosei: I agree with the issue, but I don't think a partial handwritten validator is a good tradeoff here. Full schema validation would likely mean adding a dependency such as AJV,
  which feels a bit heavy to introduce without explicit buy-in for this PR. If we want strict validation, I'd rather do that as a separate, deliberate follow-up.

src/playground/components/SettingsJsonEditorModal.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+		const timeoutId = window.setTimeout(() => {
+			setCopyStatus("idle");
+		}, 2000);
+
+		return () => {
+			window.clearTimeout(timeoutId);
+		};
+	}, [copyStatus]);
```
dyc3: I'm not a react expert, but I feel like its possible to reduce the usage of these `useEffect`s.

src/playground/components/SettingsJsonEditorModal.tsx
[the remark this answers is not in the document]
ematipico: Ignore the suggestion

src/playground/buildBiomeConfiguration.ts
```diff
@@ -0,0 +1,458 @@
...
+		},
+	};
+
+	return configuration;
+}
+
+export function parseBiomeConfiguration(
+	configuration: Configuration,
+	currentSettings: PlaygroundSettings = defaultPlaygroundState.settings,
+): PlaygroundSettings {
```
dyc3, line 382: Something I'm worried about is maintainability.  In this function is a lot of hard coded things for example. If we change configuration (add a new option, add a new variant to an option, etc.), it wouldn't necessarily be automatically be handled here.

We enjoy having codegen for this kind of tedium. Could we enhance our codegen here to make it better? I think ideally we should unify `PlaygroundSettings` with `Configuration` in some form.

This might be too big of a problem for this PR. I'm OK to defer it to a new PR, as long as we at least open an issue about it.

dyc3 (commented):
Image

can we avoid outputting the domains empty object?

src/playground/buildBiomeConfiguration.ts
mkosei: That makes sense. I’ll look into improving the codegen approach and open a new PR for it. I’ll also create an issue to track it separately.

PR #4153, opened 2026-04-14
ci: use cargo-sweep to minimize size of saved caches

dyc3:
## Summary

Our sync job takes >30 minutes, and most of that is downloading and uncompressing the build cache, and then compressing it and reuploading it. This is because the build cache can balloon to tens of gigs. Hoping this will reduce or negate that.

### DIFF
--- a/.github/actions/run-codegen/action.yaml
+++ b/.github/actions/run-codegen/action.yaml
@@ -17,7 +17,7 @@ runs:
       uses: moonrepo/setup-rust@ede6de059f8046a5e236c94046823e2af11ca670 # v1.2.2
       with:
         cache: false
-        bins: cargo-cache
+        bins: cargo-cache, cargo-sweep
 
     - name: Restore rust cache
       id: cache-restore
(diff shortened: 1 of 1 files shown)


.github/actions/run-codegen/action.yaml
```diff
@@ -29,6 +29,10 @@ runs:
           $HOME/.cargo/registry
           target/debug
 
+    - name: Stamp rust build cache
+      shell: bash
+      run: cargo sweep --stamp
```
ematipico, line 34: What does it do?

ematipico (approved):

.github/actions/run-codegen/action.yaml
dyc3: Their readme says:

> To clean everything but the latest build you will need to run it in several steps.
> 
> ```
> cargo sweep --stamp
> 
> <Insert any number of cargo build, cargo test etc...>
> 
> cargo sweep --file
> ```
> The first step generates a timestamp file which will be used to clean everything that was not used between it and the next time the file (--file) option is used.

PR #4194, opened 2026-04-27
Add Markdown playground

tidefield:
This PR adds the playground for Markdown.

Keeping in mind that the formatter pane still shows "Not supported" because biome's Configuration has markdown behind a feature. As soon as we lift or enable that feature (I don't know when's the right time), the pane should show formatting. I tested this with a local wasm build.

### DIFF
--- a/src/playground/tabs/SettingsTab.tsx
+++ b/src/playground/tabs/SettingsTab.tsx
@@ -435,6 +435,7 @@ function LanguageView({
 						[LANGUAGE.Vue]: "Vue",
 						[LANGUAGE.Svelte]: "Svelte",
 						[LANGUAGE.Astro]: "Astro",
+						[LANGUAGE.Markdown]: "Markdown",
 					}}
 					value={language ?? LANGUAGE.TSX}
 					onChangeValue={setLanguage}
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "@codemirror/lang-html": "6.4.11",
     "@codemirror/lang-javascript": "6.2.5",
     "@codemirror/lang-json": "6.0.2",
+    "@codemirror/lang-markdown": "^6.5.0",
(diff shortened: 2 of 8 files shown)


siketyan (approved):

src/playground/tabs/SettingsTab.tsx
```diff
@@ -435,6 +435,7 @@ function LanguageView({
 						[LANGUAGE.Vue]: "Vue",
 						[LANGUAGE.Svelte]: "Svelte",
 						[LANGUAGE.Astro]: "Astro",
+						[LANGUAGE.Markdown]: "Markdown",
```
siketyan, line 438: I think we should still hide the option from here if it's not usable now, as it may lead confusion for some people. Markdown can still be tested if we enter the filename manually in the multi-file mode.

dyc3:
Fyi, for dev work, if you just want to compare biome and prettier's output, you should update the prettier-compare tool that's in the main repo.

src/playground/tabs/SettingsTab.tsx
tidefield: I think it's ok because the formatter pane shows "Not supported" already. Check out the netlify link.

As I said in the PR, we will only lift the feature in the main repo when the formatter is ready for use. Until then, only the AST and CST are exposed.

Let me know if you still think we should remove this.

tidefield:
> Fyi, for dev work, if you just want to compare biome and prettier's output, you should update the prettier-compare tool that's in the main repo.

@dyc3 Thanks! I didn't know `prettier-compare` existed. I'll take a look too.

ematipico (approved):
I think the implementation is fine.

ematipico:
@tidefield please fix the CI

tidefield:
@ematipico I fixed the CI.

PR #4200, opened 2026-04-30
fix: style regression

ematipico:
## Summary

Tested locally, it works :)


src/playground/main.tsx
ematipico, line 1: This was never used

dyc3 (approved):

PR #4217, opened 2026-05-04
docs: inlineConfig example for helix

forivall:
## Summary

Add an example for `inlineConfig` for the helix editor configuration example. I'm submitting here instead of the helix wiki as the [helix wiki page](https://github.com/helix-editor/helix/wiki/Language-Server-Configurations#biome) points here for configuration.


dyc3 (approved):

ematipico:
At this point we could document all the other options. I didn't know it was possible now :)

PR #4257, opened 2026-05-18
docs(users): add Hashiverse

jimmejardine:
## Summary

Added the open-source hashiverse social network as a user of biome (https://github.com/hashiverse/hashiverse).


ematipico:
Thank you, but we don't have a presence in that social.

PR #4264, opened 2026-05-22
docs: clarify scope of `--changed` for VCS integration

### LINKED ISSUE #3133: 📚 The `defaultBranch` feature for `vcs` should specify where it is in effect
### Documentation URL

https://biomejs.dev/guides/integrate-in-vcs/#process-only-changed-files

### Description

To me it doesn't make sense that you would want this functionality for the linter, only the formatter. In many cases if you edit file A, it may break lint rules in file B that imports A.

### Expectations

The documentation and rule should update to either only apply to the formatter, or know which files downstream would also need to be checked.

### Code of Conduct

- [x] I agree to follow Biome's Code of Conduct

Dotify71:
Closes #3133.

This updates the docs for the VCS integration to clarify two things that were confusing users:
1. `--changed` applies to both the formatter and the linter.
2. It only checks the exact files that have diffs. It won't automatically trace dependencies to run full linting on downstream files that import those changed files.

This aims to address the confusion highlighted in the issue.

### DIFF
--- a/src/content/docs/guides/integrate-in-vcs.mdx
+++ b/src/content/docs/guides/integrate-in-vcs.mdx
@@ -57,7 +57,9 @@ biome check --changed
 ```
 
 :::caution
-Biome doesn't check what's changed, this means that even adding spaces or newlines to a file, will mark this file as "changed"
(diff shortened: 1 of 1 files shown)


yanthomasdev (approved):
Looks good to me, thanks!

ematipico (approved):

PR #4319, opened 2026-06-15
fix: playground syntax highlight

Netail:
## Summary

Fix playground syntax highlight, broke with the next branch merge.

Spend several hours on what the change was. As the merge included some weird downgrades in the pnpm lock, I tried `pnpm upgrade`, which did the job


ematipico:
Shouldn't we upgrade the libs in the package.json too?

ematipico (approved):

Netail:
> Shouldn't we upgrade the libs in the package.json too?

These are some underlying deps that are valid to bump due to still being in the semver range, none of the direct deps got bumped

PR #4326, opened 2026-06-18
chore: update Astro to v6.4

siketyan:
## Summary

Supersedes #4324 
Supersedes #4325 

Updated Astro and Starlight to the latest version.


siketyan:
✅ Built successfully
✅ Tables are not broken
💹 Playground works

PR #4329, opened 2026-06-20
docs: remove --colors=off from lefthook examples

Dotify71:
Resolves #4305

Removed the `--colors=off` flag from the Lefthook examples in `git-hooks.mdx` (and its translations) since recent Lefthook versions handle TTY correctly.

Also added a brief note for users on older Lefthook versions as suggested in the issue.

PR #4330, opened 2026-06-20
docs: remove --colors=off from lefthook examples

### LINKED ISSUE #4305: Remove `--colors=off` from Lefthook examples for recent Lefthook versions
## Summary
The Lefthook examples in the Git Hooks recipe currently pass `--colors=off` to Biome:

https://biomejs.dev/recipes/git-hooks/

This option was added in https://github.com/biomejs/website/pull/876 for https://github.com/biomejs/website/issues/872, where the reported reproduction used Lefthook v1.7.11.

I think this workaround may no longer be necessary for Lefthook v2.1.6+.

## Context
Lefthook has had several color / TTY related changes since v1.7.11:
- https://github.com/evilmartians/lefthook/pull/812 added better color control, including `--colors=auto|on|off`, `NO_COLOR`, `CLICOLOR_FORCE`, and propagation of those env vars to subcommands.
- https://github.com/evilmartians/lefthook/pull/1393 changed command execution so Lefthook skips PTY allocation when stdout is not a terminal.
- That fix was released in v2.1.6:
https://github.com/evilmartians/lefthook/releases/tag/v2.1.6

The PTY change seems especially relevant because the original problem was that ANSI escape sequences were emitted in an environment that did not render them. With recent Lefthook, when stdout is not a terminal, child commands should no longer see a pseudo terminal, so Biome's own color auto-detection should work correctly.

Lefthook's current docs also describe color control through `NO_COLOR` / `CLICOLOR_FORCE`:
https://lefthook.dev/configuration/colors/

## Proposal
Remove `--colors=off` from the Lefthook examples, so Biome keeps colored diagnostics in normal terminals and relies on its automatic color detection elsewhere.

Optionally, add a short note for older Lefthook versions, e.g. users who still see raw ANSI escape sequences should upgrade Lefthook or configure Lefthook colors / `NO_COLOR`.

## Motivation
I'm using Biome with Lefthook, and while checking the Git Hooks recipe I wondered why the Lefthook examples explicitly pass `--colors=off`.

I followed the blame history and found that it was added in https://github.com/biomejs/website/pull/876 for https://github.com/biomejs/website/issues/872. However, I couldn't reproduce the raw ANSI escape sequence output with the current Lefthook version(v2.1.9) in my environment, so I looked into whether Lefthook had changed its color or TTY handling since then.

## Note
I used AI assistance while researching the related Lefthook changes and drafting this issue. I have reviewed the linked issues and PRs myself, but please let me know if any of my interpretation is incorrect.

Dotify71:
Resolves #4305

As requested, this PR modifies ONLY the English  file to remove the `--colors=off` flag from the Lefthook examples and add the version compatibility note.

(Replaces closed PR #4329)

### DIFF
--- a/src/content/docs/recipes/git-hooks.mdx
+++ b/src/content/docs/recipes/git-hooks.mdx
@@ -24,7 +24,7 @@ Some examples of _Lefthook_ configurations:
     commands:
       check:
         glob: "*.{js,ts,cjs,mjs,d.cts,d.mts,jsx,tsx,json,jsonc,css}"
-        run: npx @biomejs/biome check --no-errors-on-unmatched --files-ignore-unknown=true --colors=off {staged_files}
+        run: npx @biomejs/biome check --no-errors-on-unmatched --files-ignore-unknown=true {staged_files}
   ```
 
 - Format, lint, and apply safe code fixes before committing
@@ -34,7 +34,7 @@ Some examples of _Lefthook_ configurations:
     commands:
       check:
         glob: "*.{js,ts,cjs,mjs,d.cts,d.mts,jsx,tsx,json,jsonc,css}"
-        run: npx @biomejs/biome check --write --no-errors-on-unmatched --files-ignore-unknown=true --colors=off {staged_files}
+        run: npx @biomejs/biome check --write --no-errors-on-unmatched --files-ignore-unknown=true {staged_files}
         stage_fixed: true
   ```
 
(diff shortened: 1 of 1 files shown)


src/content/docs/recipes/git-hooks.mdx
```diff
@@ -58,6 +58,10 @@ If you wish more control over which files are handled, you should use `glob`.
 
 Once configured, run `lefthook install` to set up the hooks.
 
+:::note
+If you are using an older version of Lefthook and see raw ANSI escape sequences in your terminal, you should either upgrade Lefthook or configure Lefthook colors / set `NO_COLOR`. Alternatively, you can pass `--colors=off` to Biome.
```
ematipico: Older than what? This is an important information

ematipico (changes requested):

src/content/docs/recipes/git-hooks.mdx
Dotify71: Good catch! I have updated the note to explicitly specify Lefthook `v2.1.6`, which is when the PTY allocation fix (#1393) was introduced.

ematipico (approved):

PR #4403, opened 2026-07-29
docs: document .htm support

Punisheroot:
## Summary

* update the **Language support** page to document that Biome recognizes both `.html` and `.htm` files as HTML, following the support introduced in biomejs/biome#11126

### DIFF
--- a/src/content/docs/internals/language-support.mdx
+++ b/src/content/docs/internals/language-support.mdx
@@ -31,6 +31,8 @@ Legend:
 
 *\* currently requires [explicit opt-in](/reference/configuration/#html)*
 
+Biome recognizes files with the `.html` and `.htm` extensions as HTML.
+
 ## JavaScript support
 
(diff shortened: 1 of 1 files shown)


src/content/docs/internals/language-support.mdx
```diff
@@ -31,6 +31,8 @@ Legend:
 
 *\* currently requires [explicit opt-in](/reference/configuration/#html)*
 
+Biome recognizes files with the `.html` and `.htm` extensions as HTML.
```
ematipico, line 34: Shouldn't this go below the HTML paragraph?

dyc3: There isn't a specific html section. there's stuff about super languages, but not base html

dyc3 (approved):
This is fine, I'll fix it up. I saw some other things on the language support page that need updating, so I'll take care of it.

PR #4404, opened 2026-07-30
docs: rework CLI reference

yanthomasdev:
## Summary

This PR is the docs part of https://github.com/biomejs/biome/pull/11134 with the intent of reworking the CLI reference for better readability and less redundancy.

Previously, the CLI reference was mostly automatically generated, this PR is also a proposal to go for a more manual approach, given the current format is very limiting in how we organize the information for users.


src/content/docs/reference/cli.mdx
```diff
@@ -1,1943 +1,1134 @@
...
+Defaults to `none`.
+
+#### `--log-kind=<pretty|compact|json>`
 
+For file-processing commands, selects the internal CLI log format: `pretty`, `compact`, or `json`.
 
-## Useful information
+Uses the **`BIOME_LOG_KIND`** environment variable.
 
-- When encountering symbolic links, the CLI will expand them until three levels deep. Deeper levels will result into an error diagnostic.
```
yanthomasdev, line 1930: I've decided to remove this information from here into the configuration file, felt very out of place.

Note that "error diagnostic" doesn't seem to be accurate from what I saw in the core repository; it's meant to be a warning diagnostic instead.

src/content/docs/reference/cli.mdx
```diff
@@ -1,1943 +1,1134 @@
...
+
+```shell
+biome ci --reporter=default --reporter=rdjson --reporter-file=report.json
+```
+
+## Shared flags
+
+These flags are accepted by multiple commands. Configuration flags override the equivalent settings from `biome.json` or `biome.jsonc`.
+
+<h3 id="daemon-flags">Daemon server flags</h3>
```
sarah11918: There are *so* many shared flags, that I think it's helpful to have these categories visible in the TOC for easier discovery of what exists here. Yes, it will make the TOC longer, but I think it's easier scanning and navigating!

src/content/docs/reference/cli.mdx
```diff
@@ -1,1943 +1,1134 @@
...
+
+Includes Biome rules inspired by ESLint rules in the migration.
+
+###### `--include-nursery`
+
+Includes nursery rules in the migration.
+
+## `biome search`
+
+Finds code that matches a GritQL pattern. This command is experimental.
```
sarah11918: It feels like "Experimental" should be the first thing someone sees. 

Maybe keep the original "Experimental"?

```suggestion
EXPERIMENTAL: Finds code that matches a GritQL pattern, optionally limited to a specific file path.
```

Also noting, below you mention `PATH` and `PATTERN` arguments, but I don't think these are named and defined anywhere, nor is there an example of using a path. So, maybe adding that concept here is helpful.

src/content/docs/reference/cli.mdx
```diff
@@ -1,1943 +1,1134 @@
...
+
+GritQL delimits code snippets with backticks. Because many shells interpret backticks specially, enclose the complete query in single quotes when the shell supports them.
+
+The required `PATTERN` argument is the GritQL pattern to find. It must be a search pattern because rewrites are not supported.
```
sarah11918: I wonder whether the whole use single quotes advice could be more simple, something more like this style... (Do this... to avoid...)
```suggestion
Provide the GritQL pattern to find. It must be a search pattern because rewrites are not supported. Put your GritQL query inside single quotes when the shell supports them. This avoids interpreting inline code as commands.
```
NB: I don't know what "it must be a search pattern because rewrites are not supported" means. Is there a risk of people not writing their query correctly? Are people expecting rewrites? (Maybe all that is yes, just saying that I don't know what that means or why/how much it's important, so you need to make the call.)

Also, I'm not sure of the significance of "the complete query". Does this mean, if there's a path, the path needs to be enclosed too? I think it would be assumed that "your query" (i.e. the pattern) means the COMPLETE thing you're searching for, and wouldn't need to be said. It just makes me wonder what you mean by complete. 😄

src/content/docs/reference/cli.mdx
```diff
@@ -1,1943 +1,1134 @@
...
+
+The optional `PATH` arguments limit the search to one or more files or directories. If omitted, Biome searches files in the current working directory.
```
sarah11918: This sounds great, but a) see my question above re: is this part of a "full" query? and b) can we have an example that uses a path (which would then answer a) if we could see it in the example)?

How about an example that would find all `console.log` invocation in X directory? Then we see what the full command looks like. The default, excluding it and just searching in the current working directory, is trivial by comparison.

src/content/docs/reference/cli.mdx
```diff
@@ -1,1943 +1,1134 @@
...
+
+GritQL patterns are specific to their target grammar, so a search cannot target multiple languages at once.
+
+Defaults to `javascript`.
+
+```shell
+biome search --language=javascript '`console.log($message)`'
+```
+
+## `biome explain`
```
sarah11918, line 519: Nicely done on this one!

src/content/docs/reference/cli.mdx
```diff
@@ -1,1943 +1,1134 @@
...
+#### `--formatter-enabled=<true|false>`
 
-  The file doesn't need to exist on disk, what matters is the extension of the file. Based on the extension, Biome knows how to parse the code.
+Enables or disables formatting checks for this command.
 
-  Example: ```shell echo 'let a;' | biome search '`let $var`' --stdin-file-path=file.js ```
-- **`-l`**, **`--language`**=_`ARG`_ &mdash; 
-  The language to which the pattern applies.
+```shell
+biome check --formatter-enabled=true
```
sarah11918: Just checking, isn't true the default for `check`? Is someone more likely to enable true or enable false when running this command? (goes for all of these)

src/content/docs/reference/cli.mdx
```diff
@@ -1,1943 +1,1134 @@
...
-- **`-h`**, **`--help`** &mdash; 
-  Prints help information
+```shell
+echo 'let a;' | biome check --stdin-file-path=file.js --write
+```
 
+#### `--staged`
 
-## biome clean
+Processes only staged files. This option is intended for local use.
```
sarah11918: I wonder whether "processes" is too generic here? This is about `check`ing, so even saying "Checks only staged files" feels more descriptive?

src/content/docs/reference/cli.mdx
```diff
@@ -1,1943 +1,1134 @@
...
+Enables rule profiling output. Captures timing only for rule execution, not preprocessing such as querying or building the semantic model.
 
-**Available positional items:**
-- _`NAME`_ &mdash; 
-  Single name to display documentation for.
+#### `--stdin-file-path=PATH`
 
+Reads code from standard input and writes the processed code to standard output.
 
+Biome uses `PATH` to select settings and determine the input type from its extension. Virtual paths don't need to exist or belong to the project and bypass `files.includes` and VCS ignore checks.
```
sarah11918: This might be a little *too* dense? It feels like maybe this second sentence is trying to combine too many ideas, and could be split into two for more comfortable reading.

src/content/docs/reference/cli.mdx
```diff
@@ -1,1943 +1,1134 @@
...
-  Prints help information
-
-
-## biome migrate
-
-Updates the configuration when there are breaking changes.
+Prints a troubleshooting report with CLI, platform, environment, configuration, workspace, and daemon server information.
 
-**Usage**: **`biome`** **`migrate`** \[**`--write`**\] \[_`COMMAND ...`_\]
+This command also accepts [global flags](#global-flags), the relevant [shared flags](#shared-flags), and the following additional flags:
```
sarah11918, line 87: When you say the following *relevant* shared flags, will it be obvious which of the ones in the Shared Flags section are relevant?

I wonder if, when we use this line, we should reiterate using the `help` command, like:

```suggestion
This command also accepts [global flags](#global-flags), the relevant [shared flags](#shared-flags), and the following additional flags. Use the help command (`-h`) to see a complete list.
```

src/content/docs/reference/cli.mdx
```diff
@@ -1,1943 +1,1134 @@
...
+#### `--staged`
 
-## biome clean
+Processes only staged files. This option is intended for local use.
 
-Cleans the logs emitted by the daemon.
+#### `--changed`
 
-**Usage**: **`biome`** **`clean`** 
+Processes only files changed by commits since the merge base between `REF` and `HEAD`. `REF` comes from `--since` or `vcs.defaultBranch`. Staged and unstaged changes are not included. This option is intended for CI.
```
sarah11918: This feels a bit technical/jargony, even if extremely accurate!

Maybe closer to... (make this accurate) ?

```suggestion
Processes only files with committed changes compared to your `defaultBranch` or a base reference specified with `--since`. Staged and unstaged changes are not included. This option is intended for CI.
```

src/content/docs/reference/cli.mdx
```diff
@@ -1,1943 +1,1134 @@
...
+biome explain daemon-logs
+```
+
+## `biome clean`
+
+Removes the Biome daemon server log files.
+
+## Global flags
+
+These flags are available to many commands and control general CLI behavior, diagnostics, and output.
```
sarah11918, line 536:
```suggestion
These flags are available to all commands and control general CLI behavior, diagnostics, and output.
```
Is it "all" if they are global?

sarah11918 (commented):
Hey @yanthomasdev there's some great writing and organization here! I had a few specific comments on some things, but overall I think this is very readable and user-friendly!

src/content/docs/reference/cli.mdx
yanthomasdev: I think technically some don't apply to `biome version` and `biome help`, but I don't mind being a bit innacurate in that matter if you think that's OK!

src/content/docs/reference/cli.mdx
yanthomasdev: I feel like maybe it would be repeating that a bit too much? Like, for every command basically. No strong feelings either, but I thought it was fine given the nudge to do this in the initial tip block.

src/content/docs/reference/cli.mdx
yanthomasdev: From my understanding, the GritQL syntax allows you both to search for code or to search and replace (rewrite) them, kind of like regex if you know it. I don't think people would expect rewrites to work here since the command is about *search*. 

But I guess someone could use a pattern that would rewrite the code expecting it to just work or be converted to a search one. I don't know!

yanthomasdev:
Thanks @sarah11918, applied/addressed most of your feedback (a few others I added comments to see your opinion)

ematipico:
@yanthomasdev is this safe to merge?

yanthomasdev:
Yes @ematipico, you can merge this one as well!

PR #4408, opened 2026-07-31
fix(a11y): make the homepage's scrollable code blocks keyboard accessible

AccessNexaaly:
The two code blocks in the homepage hero cannot be scrolled with a keyboard, so the lines below the fold are unreachable unless you have a mouse.

### What is happening

`src/styles/_overrides.css` gives them a fixed height and `overflow: scroll`:

```css
.linter > div > .expressive-code:nth-child(-n + 2) > figure > pre > code {
	height: 400px !important;
	overflow: scroll;
	font-size: 0.7rem;
}
```

That moves the scroll container off Expressive Code's `<pre>` and onto the `<code>` inside it. Measured on the live site:

| element | `overflow-x` | scrolls | `tabindex` |
|---|---|---|---|
| `pre` | `auto` | no (358 = 358) | none |
| `code` | `scroll` | **yes (777 vs 358)** | **none** |

So the element that actually scrolls is not focusable, and Expressive Code does not add a `tabindex` of its own (I checked `/guides/getting-started/`: 15 `<pre>` elements, none with `tabindex`; nothing overflows there, so it never shows up).

WCAG 2.1.1 Keyboard (Level A). axe reports it as `scrollable-region-focusable`, 2 nodes.

### The fix

Marks the block focusable only when it genuinely overflows, so nothing changes if the layout changes or the content gets shorter. No visual change.

### Verification

Ran axe against the live homepage, applied this exact change in the page, ran it again:

```
BEFORE: [{"id":"scrollable-region-focusable","nodes":2}]
elements given tabindex: 2
AFTER : []
```

I did not add `role="region"` and an `aria-label`. That is the fuller treatment and would announce the block as a named region, but it changes how a screen reader reads the page, so it felt like your call rather than mine. Happy to add it if you would prefer.

Found with an accessibility checker I maintain. Not selling anything here; the finding is the finding, and it is reproducible above.


yanthomasdev (commented):
Thanks @AccessNexaaly, good work here. I made a few changes to your branch to fix the issue + use ResizeObserver, I think it would be nice to add a outline for the focused code block as well for accessibility's-sake. Could you do that?

PR #4420, opened 2026-08-05
docs: rework diagnostics and env references

yanthomasdev:
## Summary

This PR is the docs part of https://github.com/biomejs/biome/pull/11248 reworking the diagnostics and environment variable reference pages for for better readability.


ematipico (approved):

codegen/src/diagnostics.rs
```diff
@@ -23,9 +23,12 @@ struct PlainDiagnostic {
 impl Default for PlainDiagnostic {
     fn default() -> Self {
         Self {
-            message: MessageAndDescription::from(markup!(
-                "This is the message of the diagnostic. It will appear in different colours based on the severity of the diagnostic."
-            ).to_owned())
+            message: MessageAndDescription::from(
+                markup!(
+                    "A diagnostic's message is displayed in a color that reflects its severity."
```
sarah11918:
```suggestion
                    "This diagnostic message is displayed in a color that reflects its severity."
```
Maybe something like this for these, @yanthomasdev ? (So as not to also lose the idea that this is a message?) Otherwise, PR approved from me!

PR #4438, opened 2026-08-16
feat: UI glow up

ematipico:
## Summary

I decided to give a UI glow-up to the website; I helped myself with a coding agent. Here's a summary:
- Added a new font for coding
- Added a new font for the whole website
- The asides have been revamped, and they are closer to Biome Diagnostics
- Removed some shadows
- Reworked the next/prev buttons at the bottom of the page
- Reworked the current page and current section style. 
- New style for the tabs
- New style for the code blocks (closer to Biome diagnostic style)
- New style for the tables
- Moved the sponsors to the right side bar so it's always visible
- The sponsors don't appear anymore on the plyagorund (they were causing issues)
- Added codspeed to platinum sponsors

Also:
- Removed old SVG we don't use anymore
- Upgraded to Biome 2.5 and enabled HTML-ish support. Found a couple of bugs I'll fix later
- Removed some old suppressions


dyc3 (approved):
I like the new font

PR #4449, opened 2026-08-17
feat: add pages for multi language rules

ematipico:
## Summary

This PR fixes a problem we've had with rules that are supported by multiple languages.

Given the following: https://biomejs.dev/linter/rules/no-excessive-lines-per-file/ , the "On this page" has a massive number of headers because we distribute the docs into multiple 

With this change, we don't have a rule page anymore, but we add `/<lang>` to the rule. Each page will have a small header to navigate to the same rule for another language. Initially I did it with a dropdown, but then we need to provide links for assistive technologies, so I went for pure links. 

I also generated redirects so that the "default" page redirects to the "language" version. 

I used a coding agent to help myself. The change is straightforward, but it touches all pages :(


nhedger (approved):
Great idea

PR #4456, opened 2026-08-20
feat: add shell installer

nhedger:
## Summary

This PR adds a shell installer for Biome so its easier to install on macOS and Linux. I'll eventually add another one for Windows. The overall installer experience was inspired from the [opencode installer](https://github.com/anomalyco/opencode/blob/dev/install), and modified to match our needs using an agent.

### Usage

```bash
# Install the latest stable version
curl -fsSL https://biomejs.dev/install | bash

# Install a specific version
curl -fsSL https://biomejs.dev/install | bash -s -- --version 2.5.9

# Force reinstall
curl -fsSL https://biomejs.dev/install | bash -s -- --force

# Install without modifying shell configuration
curl -fsSL https://biomejs.dev/install | bash -s -- --no-modify-path

# Install into a custom directory
curl -fsSL https://biomejs.dev/install | BIOME_INSTALL_DIR="$HOME/bin" bash
```

### Installation directory

This script installs Biome in the user's `$XDG_BIN_HOME` directory, or defaults to `$HOME/.local/bin`.

### DIFF
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -47,15 +47,15 @@ jobs:
       matrix:
         include:
           - platform: Linux x64
-            runner: ubuntu-latest
+            runner: depot-ubuntu-24.04
           - platform: Linux ARM64
-            runner: ubuntu-24.04-arm
+            runner: depot-ubuntu-24.04-arm
           - platform: macOS
-            runner: macos-latest
+            runner: depot-macos-latest
           - platform: macOS Intel
-            runner: macos-15-intel
+            runner: depot-macos-15-intel
           - platform: Windows x64
-            runner: windows-latest
+            runner: depot-windows-latest
           - platform: Windows ARM64
             runner: windows-11-arm
     steps:
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -53,7 +53,7 @@ jobs:
           - platform: macOS
             runner: depot-macos-latest
           - platform: macOS Intel
-            runner: depot-macos-15-intel
+            runner: macos-15-intel
           - platform: Windows x64
             runner: depot-windows-latest
           - platform: Windows ARM64
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -37,3 +37,72 @@ jobs:
 
       - name: Run tsc
         run: pnpm tsc
+
+  installers:
+    name: Installers (${{ matrix.platform }})
+    runs-on: ${{ matrix.runner }}
+    timeout-minutes: 10
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - platform: Linux x64
+            runner: ubuntu-latest
+          - platform: Linux ARM64
+            runner: ubuntu-24.04-arm
+          - platform: macOS
+            runner: macos-latest
+          - platform: macOS Intel
+            runner: macos-15-intel
+          - platform: Windows x64
+            runner: windows-latest
+          - platform: Windows ARM64
+            runner: windows-11-arm
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6.1.0
+
+      - name: Test shell installer
+        if: runner.os != 'Windows'
+        shell: bash
+        env:
+          BIOME_INSTALL_DIR: ${{ runner.temp }}/biome/bin
+          INSTALLER_HOME: ${{ runner.temp }}/biome/home
+        run: |
+          mkdir -p "$INSTALLER_HOME"
+          touch "$INSTALLER_HOME/.bashrc"
+          HOME="$INSTALLER_HOME" SHELL=/bin/bash bash public/install
+          "$BIOME_INSTALL_DIR/biome" --version
+          grep -Fq "export PATH=\"$BIOME_INSTALL_DIR:\$PATH\"" "$INSTALLER_HOME/.bashrc"
+
+      - name: Test installer with PowerShell 7
+        if: runner.os == 'Windows'
+        shell: pwsh
+        env:
+          BIOME_INSTALL_DIR: ${{ runner.temp }}\biome-pwsh
+        run: |
+          Get-Content -LiteralPath public/install.ps1 -Raw | Invoke-Expression
+          biome --version
+          if (($env:Path -split ";") -notcontains $env:BIOME_INSTALL_DIR) {
+            throw "The current process PATH was not updated"
+          }
+          $userPath = [Environment]::GetEnvironmentVariable("Path", "User")
+          if (($userPath -split ";") -notcontains $env:BIOME_INSTALL_DIR) {
+            throw "The user PATH was not updated"
+          }
+
+      - name: Test installer with Windows PowerShell 5.1
+        if: runner.os == 'Windows'
+        shell: powershell
+        env:
+          BIOME_INSTALL_DIR: ${{ runner.temp }}\biome-windows-powershell
+        run: |
+          Get-Content -LiteralPath public/install.ps1 -Raw | Invoke-Expression
+          biome --version
+          if (($env:Path -split ";") -notcontains $env:BIOME_INSTALL_DIR) {
+            throw "The current process PATH was not updated"
+          }
+          $userPath = [Environment]::GetEnvironmentVariable("Path", "User")
+          if (($userPath -split ";") -notcontains $env:BIOME_INSTALL_DIR) {
+            throw "The user PATH was not updated"
+          }
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -90,6 +90,17 @@ jobs:
           if (($userPath -split ";") -notcontains $env:BIOME_INSTALL_DIR) {
             throw "The user PATH was not updated"
           }
+          $env:Path = (($env:Path -split ";" | Where-Object { $_ -ne $env:BIOME_INSTALL_DIR }) -join ";")
+          $userPath = (($userPath -split ";" | Where-Object { $_ -ne $env:BIOME_INSTALL_DIR }) -join ";")
+          [Environment]::SetEnvironmentVariable("Path", $userPath, "User")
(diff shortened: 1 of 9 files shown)


public/install
```diff
@@ -0,0 +1,263 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Installer experience and download progress inspired by OpenCode:
+# https://github.com/anomalyco/opencode/blob/dev/install
+
+readonly APP="biome"
+readonly INSTALL_DIR="${BIOME_INSTALL_DIR:-${XDG_BIN_HOME:-$HOME/.local/bin}}"
```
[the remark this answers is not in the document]
nhedger: PR description was a mistake, we use `XDG_BIN_HOME`

public/install
```diff
@@ -0,0 +1,263 @@
...
+
+if [[ -t 2 ]]; then
+	download_with_progress "$download_url" "$tmp_dir/$APP" ||
+		fail "failed to download Biome $version; see $RELEASES_URL"
+else
+	curl --fail --location --progress-bar "$download_url" --output "$tmp_dir/$APP" ||
+		fail "failed to download Biome $version; see $RELEASES_URL"
+fi
+chmod 755 "$tmp_dir/$APP"
+mv "$tmp_dir/$APP" "$INSTALL_DIR/$APP"
```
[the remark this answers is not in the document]
nhedger: We don't have a checksum artifact to check againt

src/content/docs/guides/manual-installation.mdx
```diff
@@ -94,6 +94,23 @@ docker run -v $(pwd):/code ghcr.io/biomejs/biome format --write
 
 ## Using a published binary
 
+The installer downloads the latest stable binary for macOS or Linux to `${XDG_BIN_HOME:-$HOME/.local/bin}` and adds it to your `PATH`:
```
ematipico: I don't think the wording is clear enough to state that Windows isn't supported by the installer

ematipico (commented):
What happens if the installer is executed on Windows?

src/content/docs/guides/manual-installation.mdx
nhedger: Fair enough. I moved the instructions under a `Linux and macOS` heading which should make it clearer

ematipico (commented):
Looks good to me! However, I feel like this is `next` material. It's a new feature, and we could add it to the blog post, and update READMEs or other sections of the site.

nhedger:
Technically this is release agnostic, but it see what you mean. Would you prefer i target `next` then ?

ematipico:
I think we have a good opportunity to have it in our blog post, and perhaps update the homepage with it

nhedger:
I'll squeeze the windows installer in there, and move to `next` as the target

nhedger:
Added PowerShell installer, updated the docs and added workflows to test the installers on common platforms/architectures

public/install
```diff
@@ -0,0 +1,263 @@
...
+	if [[ "$installed_version" == "$version" ]]; then
+		printf "\n${BLUE}Biome version ${NC}%s${BLUE} is already installed at ${NC}%s\n\n" \
+			"$version" "$INSTALL_DIR/$APP"
+		exit 0
+	fi
+fi
+
+mkdir -p "$INSTALL_DIR"
+tmp_dir="$(mktemp -d "${TMPDIR:-/tmp}/biome-install.XXXXXX")"
+trap 'rm -rf "$tmp_dir"' EXIT HUP INT TERM
```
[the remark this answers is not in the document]
nhedger: Addressed in f7d44968 by staging the download inside the installation directory, keeping the final replacement on the same filesystem. The same adjustment was applied to the PowerShell installer.

public/install.ps1
```diff
@@ -0,0 +1,216 @@
...
+$DownloadUrl = "$ReleasesUrl/download/$ReleaseTag/$Filename"
+$BinaryPath = Join-Path $InstallDir "$App.exe"
+
+if (-not $Force -and (Test-Path -LiteralPath $BinaryPath -PathType Leaf)) {
+	try {
+		$InstalledVersion = ((& $BinaryPath --version 2>$null) | Out-String).Trim()
+		$InstalledVersion = $InstalledVersion -replace "^Version:\s*", ""
+		if ($InstalledVersion -eq $Version) {
+			Write-Host "`nBiome version $Version is already installed at $BinaryPath`n" -ForegroundColor Blue
+			return
```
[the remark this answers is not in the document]
nhedger: Addressed in f7d44968. Matching-version reruns now restore both PATH entries unless -NoModifyPath is set, and the Windows integration tests remove both entries and verify that a rerun restores them.

.github/workflows/integrate.yaml
```diff
@@ -37,3 +37,94 @@ jobs:
...
+          - platform: Linux ARM64
+            runner: ubuntu-24.04-arm
+          - platform: macOS
+            runner: macos-latest
+          - platform: macOS Intel
+            runner: macos-15-intel
+          - platform: Windows x64
+            runner: windows-latest
+          - platform: Windows ARM64
+            runner: windows-11-arm
```
ematipico, line 60: We should use depot runners

ematipico (commented):
We could update the getting started guide to include this command (`curl` tab maybe). I updated https://github.com/biomejs/website/pull/4472 so that the installation command is in the homepage too

.github/workflows/integrate.yaml
nhedger: Moved all runners to depot, except Windows ARM because depot doesnt have ARM runners for Windows.

nhedger:
> We could update the getting started guide to include this command (`curl` tab maybe). I updated #4472 so that the installation command is in the homepage too

Added the standalone command to the PackageManagerCommand component, but since your change targeted main and this PR targets next, there wasn't the array stuff.

nhedger:
Switched to a GitHub Actions runner for `macos-15-intel`, because depot doesn't have intel-based macOS runners.

.github/workflows/integrate.yaml
ematipico: We do use depot for all releases in Biome


https://github.com/biomejs/biome/blob/main/.github%2Fworkflows%2Frelease.yml#L119-L136

This means that we're testing the installation for something we don't support maybe?

nhedger: We already build an ARM64 binary for Windows using cross-compilation. However, this PR requires a native Windows ARM64 runner to verify that the shell script fetches the correct binary for the host system.

Same reason why i just changed switched to GitHub actions for `macos-15-intel`.

src/content/docs/en/guides/manual-installation.mdx
```diff
@@ -96,6 +96,43 @@ docker run -v $(pwd):/code ghcr.io/biomejs/biome format --write
...
 ## Using a published binary
 
+### Linux and macOS
+
+The installer downloads the latest stable binary to `${XDG_BIN_HOME:-$HOME/.local/bin}` and adds it to your `PATH`:
+
+```shell
+curl -fsSL https://biomejs.dev/install | bash
+```
+
```
yanthomasdev, line 106: Since its adding the Biome binary to your `PATH`, I assume we should ask users to restart their shell so it picks it up? IIRC Linux/MacOS don't automatically update the session's PATH like PowerShell does.
```suggestion

:::tip
You might need to restart your shell after installation before you can run `biome`.
:::

```

public/install
```diff
@@ -0,0 +1,263 @@
...
+
+Options:
+    -h, --help              Display this help message
+    -v, --version <version> Install a specific version
+    -f, --force             Download even if the version is already installed
+        --no-modify-path    Don't modify shell config files
+
+Environment variables:
+    BIOME_INSTALL_DIR       Installation directory
+    XDG_BIN_HOME            Default installation directory (~/.local/bin)
```
yanthomasdev: One enviroment variable being "Installation directory" and the other being "Default installation directory" feels a bit confusing and could make people wonder "what should I actually set?". From what I know, `XDG_BIN_HOME` is not really a specific Biome installer env variable but instead something from Unix systems, right? I feel like we could instead only display `BIOME_INSTALL_DIR` here and document it will fallback to the value of `XDG_BIN_HOME` when unset.


```suggestion
    BIOME_INSTALL_DIR       Set the installation directory
```

public/install
```diff
@@ -0,0 +1,263 @@
...
+    BIOME_VERSION           Version to install
```
yanthomasdev:
```suggestion
    BIOME_VERSION           Set the version to install
```

public/install
```diff
@@ -0,0 +1,263 @@
...
+usage() {
+	cat <<'EOF'
+Biome Installer
+
+Usage: install [options]
+
+Options:
+    -h, --help              Display this help message
+    -v, --version <version> Install a specific version
+    -f, --force             Download even if the version is already installed
```
yanthomasdev: I would rephrase it to focus on the idea that you'd use the flag to reinstall with a fresh download (if you have any corruption issues, etc.) which I think is the main goal here?
```suggestion
    -f, --force             Force a fresh download and reinstallation
```

public/install
```diff
@@ -0,0 +1,263 @@
...
+        --no-modify-path    Don't modify shell config files
```
yanthomasdev: Starting with "Don't" here instead of a verb like the other descriptions feels a bit weird, what about something like this? That would also be more explicit about what it's doing, "modify shell config files" might imply it's doing a lot more than it actually is.
```suggestion
        --no-modify-path    Skip adding Biome to PATH
```

yanthomasdev (commented):
Good work @nhedger, I have a few suggestions. I've made a few suggestions only in the Unix script, but they are also meant for the PowerShell one.

src/content/docs/en/guides/manual-installation.mdx
nhedger: Good point

public/install
nhedger: Much better indeed, thanks

ematipico (approved):
Reminder to us that we need add this to the v2.6 blog post

yanthomasdev (approved):

PR #4460, opened 2026-08-20
feat(playground): overhaul playground UI

dyc3:
## Summary

Redesigns the playground

- removes the concept of single file mode
- adds biome config "ejection" - where visual settings are replaced with an actual biome.json entry
- moves debug tools to the right side as expandable panels, allowing you to view output and debug tools simultaneously
- adds a toggle for comparing prettier

I am not very good at react, so my clanker may have done some crazy things. Fable 5 did most of the heavy lifting.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


dyc3:
it appears netlify doesn't do preview builds for stacked prs...

ematipico:
@dyc3  No, that's not it. It does previews only for PRs that target specific branches. `main` and `next` are, in fact, configured to generate previews because they can be deployed.

If a branch can't be deployed, PRs against it won't generate previews. In your case you would need to add your branches as deployable. 

Once this PR will target `main`, you'll see the preview.

ematipico:
I got some initial feedback (external, from a person who doesn't know these things):
- They like the side-by-side comparison with Prettier. Easier to read
- Moving the IR, Syntax, etc. to the side doesn't work at all. Hard to spot and to read too; you need to bend your head. They prefer the current design

dyc3:
Hmm, I'll have to cook something else up then

ematipico:
Some thoughts:

In the playground, we essentially have three macro views: options, user-facing views (console, diagnostics, fixes, format, etc.) and developer-facing views (IR, syntax, etc.). So maybe we could add a switch that flips between the user-facing and developer-facing views.

dyc3:
I put our internal tools behind a collapsible toggle. Arguably, syntax is useful for users for looking up names of syntax nodes for grit queries, so maybe we could have that one more discoverable

ematipico:
I prefer the current version because each view has its own space. With this version, you can end up with all the views open and see nothing because it's all [amassed](https://deploy-preview-4460--biomejs.netlify.app/playground/?view=formatter-ir%2Csyntax%2Ccontrol-flow%2Csemantic-model%2Ctypes-ir%2Ctypes-registered). If I toggle the Syntax view, I want to see **only** that view

dyc3:
I think the multiple tools is useful. Sometimes I want to look at both the syntax and formatter IR, for example. Or you might want to look at the types IR and registered at the same time. The flexibility is nice, toggling too many on is user error, IMO.

I made switching the default behavior, and put the multiple tools behavior behind shift-clicking.

ematipico:
That's fair.

I would like to hide the user view if possible. Sometimes, to me, it's useless and I want to look at the syntax tree with as much space as possible

dyc3:
You mean hide the output? or the left sidebar?

ematipico:
Some feedback:
- I would move the editor options back at the top. Changing language is the most used thing (at least for me), so I wouldn't want to scroll all the way down
- Clicking `edit root config` was a bit disruptive. Sidebar changed, source code changed, etc. I suppose it's normal but I didn't expect it. It took me a while to understand how to go back (delete button)
- Mobile needs some polishing. There are a few gaps. Check the screenshots, the top part of the sidebar has some gaps at the top. Also, I would make them **full screen**, there's no reason in leaving some space **on the side**. There's already a small space, so let's use all of it.


deploy-preview-4460--biomejs netlify app_playground__view=types-registered format=false prettier=true

deploy-preview-4460--biomejs netlify app_playground__view=types-registered format=false prettier=true (1)

dyc3:
Tweaked the copy for the buttons and polished the spacing

ematipico:
More defects found on mobile 

Uneven separation of buttons (up a bottom)



Weird gap at the bottom, probably the sidebar doesn't fully expand 



Buttons X aren't aligned, and maybe too small

ematipico:
"Internals" still keep a number, which is now not needed since we can only select one at the time

Screenshot 2026-09-06 at 09 59 26

The arrows here are different, and they don't work. It's for collapsing diagnostics/console.

https://github.com/user-attachments/assets/822519a5-82e9-4960-8f9d-d0ea34389bf3

The dropdowns have different styles. The first two have brighter colours
Screenshot 2026-09-06 at 10 00 34

On mobile, when opening an internal, there's still a black gap. It seems it doesn't fit the full height.
Screenshot 2026-09-06 at 10 01 48

dyc3:
That thing with the arrows doesn't happen for me, that's so weird. I guess I need to swap them out for actual icons, iirc they're just characters right now

ematipico:
I use Firefox

ematipico:
The diagnostics/console window doesn't collapse when you click the arrow. I would expect it to collapse to the bottom.

dyc3:
Strange, it works for me...

edit: nvm, found the bug i think

ematipico (approved):

ematipico:
@dyc3 the "Unstable" label was lost. When selecting the language, YAML, Markdown and SASS had `(Unstable)` next to them.

Can you fix that please?

dyc3:
@ematipico yes, will do

PR #4466, opened 2026-08-21
fix: preserve generated diagnostic markup

### LINKED ISSUE #1596: 🐛 Documentation: smart quotes conversion is applied to diagnostic code blocks
### Environment information

```block
Not applicable (bug applies to the biomejs.dev website)
```

### What happened?

On the documentation website, smart quotes substitution is currently applied to diagnostic code blocks. This means that for instance, a `"` character (U+0022 QUOTATION MARK / `&quot;`) is displayed as `“` (U+201C FT DOUBLE QUOTATION MARK `&ldquo;`). Sometimes, the character is also displayed as an HTML entity code (“`&quot;`” instead of `"`).

Example from the [`useGoogleFontPreconnect`](https://biomejs.dev/linter/rules/use-google-font-preconnect/#_top) page:

### Expected result

* Access https://biomejs.dev/linter/rules/use-google-font-preconnect/#_top
* The diagnostic code block is formatted as:

```
  ⚠ The attribute rel=“preconnect” is missing from the Google Font.
  
  > 1 │ <link href=“https://fonts.gstatic.com&quot;/>
      │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    2 │ 
```

instead of

```
  ⚠ The attribute rel="preconnect" is missing from the Google Font.
  
  > 1 │ <link href="https://fonts.gstatic.com"/>
      │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    2 │ 
```

### Code of Conduct

nhedger:
## Summary

This PR fixes #1596 by rendering generated diagnostics through an Astro component, preventing smart punctuation from altering diagnostic output while preserving its formatting.

Also regenerates the rule and assist documentation.

Used an agent to investigate and implement the fix

### Before

Screenshot 2026-08-21 at 09 43 50

### After

Screenshot 2026-08-21 at 09 43 41


codegen/src/lintdoc.rs
```diff
@@ -1355,13 +1367,18 @@ fn write_documentation(
                             &mut diagnostics_writer,
                         )
                         .context("To print diagnostics or actions")?;
+                        write!(diagnostic_html, "</code></pre>")?;
+                        writeln!(
+                            content,
+                            "<RawDiagnostic html={{{}}} />",
+                            serde_json::to_string(&String::from_utf8(diagnostic_html)?)?
```
ematipico, line 1369: Why do we need serde? I think `String::from_utf8` is already enough

src/components/RawDiagnostic.astro
ematipico, line 1: Do we really need a component? Can't we write `<Fragment set:html={html} />` directly in the codegen?

codegen/src/lintdoc.rs
nhedger: `String::from_utf8` gives us the HTML text, but `set:html` expects that text as a quoted JavaScript string inside `{}`. `serde_json::to_string` adds the quotes and escapes any quotes in HTML attributes, such as `class="language-text"`

src/components/RawDiagnostic.astro
nhedger: Good point, i'll address that

ematipico (approved):
Great fix! Simpler than I thought

Nicolapps:
Thanks!

PR #4467, opened 2026-08-21
fix: hide language picker on English-only pages

### LINKED ISSUE #4437: bug: Language picker on the Playground navigates to a 404 page
### What happened

On the Playground page, selecting a non-English language from the language
picker navigates to a localized URL that does not exist, resulting in a 404.

The Playground is an English-only page: `src/pages/playground.astro` renders a
Starlight page with `template: "splash"` under `defaultLocale: "root"`, so there
is no localized variant for any of the languages offered by the picker.

### Steps to reproduce

1. Open https://biomejs.dev/playground/
2. Click the language picker in the top-right corner of the header.
3. Select any non-English language (e.g. 简体中文).
4. The browser navigates to https://biomejs.dev/zh-cn/playground/ and shows a
   404 page.

### Expected behavior

On pages that only exist in the default locale, the language picker should not
link to a non-existent localized URL. Either hide the picker on those pages, or
keep the user on the current page (e.g. `/playground/`) instead of navigating
to a 404.

### Actual behavior

The picker rewrites the current path with the selected locale prefix
(e.g. `/playground/` → `/zh-cn/playground/`), producing a 404. This comes from
Starlight's default `LanguageSelect`, which generates a locale-prefixed URL even
when the page has no localized version.

### Additional context

- The same issue also affects `/enterprise/`:
  https://biomejs.dev/enterprise/ returns 200, while
  https://biomejs.dev/zh-cn/enterprise/ returns 404.
- The Playground page also emits broken `hreflang` alternate links pointing to
  the same non-existent pages:

  ```html
  <link rel="alternate" hreflang="zh-CN" href="https://biomejs.dev/zh-cn/playground/" />
  ```

### Screenshots
Image

nhedger:
## Summary

This PR fixes #4437 by hiding the language picker and removing invalid `hreflang` links on English-only pages using declarative `localized` metadata.

The new `localized` metadata defaults to `true`. New English-only pages should set `localized: false` in their Starlight frontmatter.

Used an agent to investigate and implement the fix


src/pages/enterprise.astro
```diff
@@ -9,6 +9,7 @@ import StarlightPage from "@astrojs/starlight/components/StarlightPage.astro";
             description: "Commercial support for Biome",
             template: "splash",
             editUrl: false,
+			      localized: false,
```
ematipico: mixed tabs with spaces (should be spaces)

ematipico (approved):
Technically, we should add the `localized` to all our code-generated pages (actions, rules, cli, etc.), but that can be done later

PR #4498, opened 2026-08-31
feat: directory Biome plugins

ematipico:
## Summary

I will make sure to empty the content collection before merging. For now, let's keep it like this so that we can see how the data is rendered

### DIFF
--- a/src/content/docs/en/community-plugins.mdx
+++ b/src/content/docs/en/community-plugins.mdx
@@ -1,0 +1,28 @@
+---
+title: Community plugins
+description: A list of plugins maintained by the community
+---
+import ListPlugins from "@/components/ListPlugins.astro";
+
+This is a list of plugins created and maintained by the community.
+
+<ListPlugins />
+
+## How to add your plugin
+
+If you want to add your plugin to the list:
+1. Open the <a href="https://github.com/biomejs/website/edit/main/src/content/docs/content/plugins.json" target="_blank" rel="noopener">Content Collection file</a>
+2. Add a new entry to the list. We require the following information for each package:
+   - `id`: It's a unique identifier, so make sure there isn't another in the list. Examples: "biome-plugin-foobar`, "Superfast config",
+   - `url`: A URL of your plugin. It can be a website or the URL to the source code.
+   - `packages`: One or more packages (in case your repository/website ships more than one package). Each entry is an object with:
+      - `name`: The name of the package published on npm. Examples: `biome-plugin-foobar`, `@org/biome-config-fast`.
+      - `description`: A brief explanation of the package.
+3. Send a PR. A member of the team will merge it.
+
+
+## What this page isn't
+
+This isn't a repository or registry for plugins. The team doesn't take responsibility for the contents of the plugins the users decide the list on this page, or whether they are deprecated.
+
+
--- a/astro.config.ts
+++ b/astro.config.ts
@@ -349,6 +349,14 @@ export default defineConfig({
 					],
 				},
 				{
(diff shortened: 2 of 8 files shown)


Netail:
Perhaps a reference to `How to add your plugin` section from the publish packages page? (Previous PR)

ematipico:
> Perhaps a reference to `How to add your plugin` section from the publish packages page? (Previous PR)

I'll do it later, toward the end. I wanted to have some feedback regarding the directory itself, if that could work.

Netail:
For an initial setup, looks fine to me. I don't think a community plugins package doesn't have to be too big. Perhaps the pertial repo URL (e.g. `Netail/my-awesome-plugins`) as the identifier? Altho could be complicated with other platforms

ematipico:
> For an initial setup, looks fine to me. I don't think a community plugins package doesn't have to be too big. Perhaps the pertial repo URL (e.g. `Netail/my-awesome-plugins`) as the identifier? Altho could be complicated with other platforms

That's because you assume there will be hosted on github. They could be hosted anywhere, that's why I left a full fledged URL

Netail:
> > For an initial setup, looks fine to me. I don't think a community plugins package doesn't have to be too big. Perhaps the pertial repo URL (e.g. `Netail/my-awesome-plugins`) as the identifier? Altho could be complicated with other platforms
> 
> That's because you assume there will be hosted on github. They could be hosted anywhere, that's why I left a full fledged URL 

I mean per default suggested in the docs for filling in the id, still open an open field up. (Not that we do the parsing) But if hosted on github suggest using that as id. Codeberg & gitlab would also match something similar.

ematipico:
I am confused now, and I think I misunderstood what you were suggesting. Can you be clearer, please?

src/content/docs/en/community-plugins.mdx
```diff
@@ -0,0 +1,28 @@
...
+This is a list of plugins created and maintained by the community.
+
+<ListPlugins />
+
+## How to add your plugin
+
+If you want to add your plugin to the list:
+1. Open the <a href="https://github.com/biomejs/website/edit/main/src/content/docs/content/plugins.json" target="_blank" rel="noopener">Content Collection file</a>
+2. Add a new entry to the list. We require the following information for each package:
+   - `id`: It's a unique identifier, so make sure there isn't another in the list. Examples: "biome-plugin-foobar`, "Superfast config",
```
Netail, line 16: If we should suggest users to fill in a format of `Netail/my-awesome-plugins`. 
Like in github, codeberg & gitlab urls, but manually filled in, not parsed from the url value, so it's still open for interpretation & customizable

The current examples feel more like values for the "name" field (package names). And to keep a certain standard in the id, the values might go all kind of directions

ematipico: Ok I see what you mean. There are two problems:
- we need an `id` because that's how content collections work. I left open ended. We could enforce it. Initially used the repository, but I wanted to let users provide a "marketing" name
- we can't use the package name as identifier because a repository might contain multiple packages 


Maybe we can flip it. Use the npm package.json name as identifier. The repository becomes metadata. If two packages are in the same repo, we'll have two entries with the same site.

What do you think?

Netail: Sounds good to me

ematipico:
@Netail added the keyword part [`7f94e57` (this PR)](https://github.com/biomejs/website/pull/4498/commits/7f94e57e067cd2dc7cdf36db6e1edccf2d403e87)

Netail (approved):

PR #4504, opened 2026-09-03
fix(website): keep version select and search visible on narrow screens

ignaciobockl:
## Summary

On notebook/small-desktop widths (roughly 800px–1280px) the top navigation bar overflows: the .right-group (social icons + theme + language + version selects) doesn't fit next to the site title and search box, so the version select (2.x) gets clipped and an horizontal scroll appears.

- Added src/styles/_header.css to collapse the header responsively:
  - 800px–1280px: trim social icons to the first three and tighten the gap.
  - 800px–1152px: reduce the title to just the logo so the search box keeps its space.
  - 800px–864px: hide social icons entirely (single-line layout still overhangs at the exact breakpoint).
- Added a MobileMenuFooter override (src/components/starlight/MobileMenuFooter.astro) and registered it in stro.config.ts so the version select stays reachable from the mobile menu.

Verified the version select stays visible and there is no horizontal scroll at 360 / 768 / 800 / 850 / 900 / 1024 / 1182 / 1280 / 1366px, and pnpm tsc + pnpm exec biome check pass.

### DIFF
--- a/astro.config.ts
+++ b/astro.config.ts
@@ -1072,6 +1072,7 @@ export default defineConfig({
 				SiteTitle: "./src/components/starlight/SiteTitle.astro",
 				Hero: "./src/components/starlight/Hero.astro",
 				LanguageSelect: "./src/components/starlight/LanguageSelect.astro",
+				MobileMenuFooter: "./src/components/starlight/MobileMenuFooter.astro",
 				Footer: "./src/components/starlight/Footer.astro",
 				PageSidebar: "./src/components/starlight/PageSidebar.astro",
 			},
--- a/astro.config.ts
+++ b/astro.config.ts
@@ -1069,6 +1069,7 @@ export default defineConfig({
 			},
 			components: {
 				Head: "./src/components/starlight/Head.astro",
+				Header: "./src/components/starlight/Header.astro",
 				SiteTitle: "./src/components/starlight/SiteTitle.astro",
 				Hero: "./src/components/starlight/Hero.astro",
 				LanguageSelect: "./src/components/starlight/LanguageSelect.astro",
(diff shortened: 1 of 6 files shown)


ematipico:
Any chance to have some screenshots? Before and after

ignaciobockl:
Thanks for the review, @ematipico ! Here are before/after screenshots:

**1024px**
| Before | After |
|---|---|
| before-1024 | after-1024 |

**1182px — reported notebook width**
| Before | After |
|---|---|
| before-1182 | after-1182 |

*Before: version select (`v2.x`) is clipped outside the viewport. After: fully visible with no horizontal scroll.*

**1366px**
| Before | After |
|---|---|
| before-1366 | after-1366 |

*At 1366px all 7 social icons are kept. The header stays unchanged.*

**Note on social icons:** between 800–1280px the fix trims the social icons (shows 3/7) and at 800–864px hides them to keep `Search` + `English` + `v2.x` visible without overflow. If you prefer to keep all socials, I can add a collapsible/overflow menu (`⋯` button) for the extra icons instead — let me know and I'll push it as a follow-up change.

ignaciobockl:
Thanks for the review, @ematipico! The reported bug is now resolved — updated screenshots with the final implementation (functional ⋯ overflow menu instead of the CSS-only trim):

**1024px**

| Before | After |
|---|---|
| before-1024 | after-1024 |

**1182px — reported notebook width**

| Before | After |
|---|---|
| before-1182 | after-1182 |

Before: version select (v2.x) is clipped outside the viewport. After: fully visible with no horizontal scroll.

**1366px**

| Before | After |
|---|---|
| before-1366 | after-1366 |

At 1366px all 7 social icons are kept. The header stays unchanged.

**1182px**

|  ⋯ menu open |
|---|
| after-1182-panel-open |

The 4 remaining icons are reachable from the ⋯ dropdown (closes with outside click).

ignaciobockl:
Hello @ematipico , good day! I look forward to your review, thank you.

ematipico:
I actually preferred the before (even if has some defects) because it still provide the most important links (docs, playground, etc.). With this change, they are lost.

ematipico:
Isn't there a solution where we **don't** override so many Starlight styles? Instead, why don't we use our own styles?

PR #4508, opened 2026-09-05
docs: rework configuration reference page

### LINKED ISSUE #4485: 📚 Domains not mentioned on Configuration page
### Documentation URL

https://biomejs.dev/reference/configuration/

### Description

linter.domains config options missing, while mentioned here https://biomejs.dev/linter/domains/

### Expectations

Mention domains config options on Configuration page too

### Code of Conduct

- [x] I agree to follow Biome's Code of Conduct

### LINKED ISSUE #4486: 📚 Update version number of $schema
### Documentation URL

https://biomejs.dev/reference/configuration/ (2.3.11)
https://biomejs.dev/guides/configure-biome/#configuration-file-structure (2.4.13)
 
### Description

New users will most likely install the latest versions, so it would be nice to update the remote $schema url with the latest version

### Expectations

Keep the version in the remote $schema url in sync with the latest version

Just like on https://biomejs.dev/guides/manual-installation/

### Code of Conduct

- [x] I agree to follow Biome's Code of Conduct

yanthomasdev:
## Summary

This pull request completely reworks the Configuration page with the goal to make it more approachable, readable, and consistent.

- Closes #4485
- Closes #4486

Note that I've added reserved options to the page like `css.globals`, let me know if we'd rather not documented those.

The changes in this PR are also being added to the core repository: https://github.com/biomejs/biome/pull/11629


src/content/docs/en/guides/configure-biome.mdx
```diff
@@ -27,20 +28,18 @@ provides. At the moment, Biome provides three tools: the formatter, the linter
...
-  "linter": {
-    "enabled": false
-  },
-  "assist": {
-    "enabled": false
-  }
-}
-```
+<Code title="biome.json" lang="json" code={`{
+  "$schema": "https://biomejs.dev/schemas/${version}/schema.json",
```
dyc3, line 32: is the switch from markdown syntax to a component necessary because of the dynamic version?

src/content/docs/en/reference/configuration.mdx
```diff
@@ -1,190 +1,152 @@
...
+Biome will also warn you when your `$schema` URL version doesn't match the package version.
+:::
+
+## `extends`
+
+**Type:** `"//" | string[]`
+
+Extends the current configuration with settings from one or more other Biome configurations.
+
+In a nested configuration, use `"//"` to extend the root configuration without specifying a relative path. This works at any nesting depth:
```
dyc3, line 56: Is "nested configuration" clear enough here? Maybe we should link to monorepo docs to elaborate.

src/content/docs/en/reference/configuration.mdx
```diff
@@ -1,190 +1,152 @@
...
+```json title="biome.json"
+{
+  "extends": ["./common.json", "@org/shared-configs/biome"]
+}
+```
 
-The order of paths to extend goes from least relevant to most relevant.
+Biome merges configurations from left to right. For options that accept a single value, later configurations take precedence. For options that accept a list, Biome combines their entries. The current file is applied last.
 
-Since v2, this option accepts a string that must match the value `"//"`, which can be used
```
dyc3, line 41: even though v2 has been out for a while, I don't want to lose the fact that `//` became available in v2. some users are still on v1

dyc3 (commented):
So far, this is definitely the right direction. I didn't get very far since the diff is big, I'll try to finish reviewing it tomorrow.

src/content/docs/en/reference/configuration.mdx
```diff
@@ -194,86 +156,69 @@ the following configuration:
...
 
-:::caution
-This option has been **deprecated** and may be removed in a future release.
-Please use the
-[_force-ignore syntax_](/reference/configuration/#interaction-with-the-scanner)
-with `files.includes` instead.
+**Type:** `string[]`
+
+:::caution[Deprecated]
+This option is deprecated and may be removed in a future release. Use [force-ignore patterns](#glob-syntax-reference) in `files.includes` instead, or run `biome migrate` to update the configuration.
```
dyc3:
```suggestion
This option is deprecated and will be removed in a future release. Use [force-ignore patterns](#glob-syntax-reference) in `files.includes` instead, or run `biome migrate` to update the configuration.
```

dyc3 (approved):
Yeah, looks great! just nitpicks mostly

src/content/docs/en/guides/configure-biome.mdx
yanthomasdev: Yes, otherwise it would be rendered literally as `${version}`

src/content/docs/en/reference/configuration.mdx
yanthomasdev: We link to the monorepo docs below, do you think we should also link it in this sentence?


```suggestion
In a [nested configuration](/guides/big-projects/#monorepo), use `"//"` to extend the root configuration without specifying a relative path. This works at any nesting depth:
```

src/content/docs/en/reference/configuration.mdx
yanthomasdev: My thought process for this is to generally treat things added in the major version as status quo and only mention by version things that were added in minor versions from that same major, so things added in Biome 2.1, 2.2, etc.

We have a v1.x site with its own configuration reference intact, so I think we should recommend users to go there for precise information, instead of trying to find previous behavior in newer majors references.

src/content/docs/en/reference/configuration.mdx
dyc3: That would help. But the thing is that `extends` does one thing, but it covers 2 different use cases:
- extend the root config
- extend configs from other packages

My concern is more that "nested configuration" leans more toward jargon, particularly for novice users. What about something like:

> When a subdirectory has its own Biome configuration (a nested configuration), use "//" to extend the configuration at the project root ...

src/content/docs/en/reference/configuration.mdx
ematipico: > even though v2 has been out for a while, I don't want to lose the fact that // became available in v2. some users are still on v1

True, but when we launched v2, we assumed we write docs for v2. That's why we have the v1 dropdown

src/content/docs/en/reference/configuration.mdx
```diff
@@ -286,129 +231,97 @@ For example, if your Biome configuration is located at `frontend/biome.json` and
...
+- `style`: Enforces consistent and idiomatic code.
+- `suspicious`: Detects code that is likely to be incorrect or unintended.
 
-Each group can accept, as a value, a string that represents the severity or an object where each rule can be configured.
+A group accepts either a string level or an object. A string applies one level to the group:
 
-When passing the severity, you can control the severity emitted by all the rules that belong to a group.
-For example, you can configure the `a11y` group to emit information diagnostics:
+- `"off"`: Disables the group.
+- `"on"`: Enables the group using each rule's default severity. Recommended rules default to error severity, while other rules default to warning severity.
```
ematipico:
```suggestion
- `"on"`: Enables the group using each rule's default severity.
```

The first sentence contradicts the second sentence. The second sentence isn't true anymore; it's a leftover from v1.

src/content/docs/en/reference/configuration.mdx
```diff
@@ -420,580 +333,699 @@ For example, you can configure the `a11y` group to emit information diagnostics:
...
+**Type:** `boolean`
+
+Enables or disables the recommended rules in a non-nursery group.
+
+The group's `preset` setting takes precedence when both options are set.
+
+### `linter.rules.[group].[rule]`
 
-Example:
+**Type:** `"off" | "on" | "info" | "warn" | "error" | { level: "off" | "on" | "info" | "warn" | "error", options?: object, fix?: "none" | "safe" | "unsafe" }`
```
ematipico, line 358: nit: Wouldn't this be more readable on multiple lines?

src/content/docs/en/reference/configuration.mdx
```diff
@@ -420,580 +333,699 @@ For example, you can configure the `a11y` group to emit information diagnostics:
...
+
+Configures an individual lint rule.
+
+A rule accepts either a string level or an object. A string applies one level to the rule:
+
+- `"off"`: Disables the rule.
+- `"on"`: Enables the rule using its default severity. Recommended rules default to error severity, while other rules default to warning severity.
```
ematipico:
```suggestion
- `"on"`: Enables the rule using its default severity.
```

src/content/docs/en/reference/configuration.mdx
```diff
@@ -1005,913 +1037,627 @@ A list of global names that Biome should ignore (analyzer, linter, etc.)
...
 
-:::danger
-Setting this option to `false` is **highly discouraged** because it could cause many problems with other tools:
-- https://thoughtbot.com/blog/no-newline-at-end-of-file
-- https://callmeryan.medium.com/no-newline-at-end-of-file-navigating-gits-warning-for-android-developers-af14e73dd804
-- https://unix.stackexchange.com/questions/345548/how-to-cat-files-together-adding-missing-newlines-at-end-of-some-files
+Controls whether formatted files end with a newline.
 
-Disable trailing newline at your own risk.
+:::caution[Keep trailing newlines enabled]
```
ematipico, line 1186: Q: why the change from danger to caution?

ematipico (commented):
I spot some info that is stale about `"on"`

src/content/docs/en/reference/configuration.mdx
yanthomasdev: You mean adding more than one Type paragraph? I mostly try to keep these simple so there's not a lot of thinking when adding those and rather focus on showing the syntax in code samples below.

Not against doing it, though!

src/content/docs/en/reference/configuration.mdx
yanthomasdev: I basically copied the same aside I updated from the CLI page which is also a caution. I generally use caution as "you probably don't want to do this" and danger as "you DEFINITELY don't want to do this", no strong feelings about it if you feel it's better as danger, but we would need to update the other pages with the same warning as well for consistency's sake.

src/content/docs/en/reference/configuration.mdx
ematipico: Just thinking out loud. There are many variants, but I also don't have strong preferences, so let's keep it like this for now

src/content/docs/en/reference/configuration.mdx
ematipico: Oh ok. Thank you!

ematipico (approved):
Awesome rework, thank you @yanthomasdev !

PR #4520, opened 2026-09-09
Clarify the usage of 'use strict' in ESM and CommonJS files.

alexrudd2:
## Summary

The default assumption of strict mode _without_ declaring `"type": "module"` in `package.json` came as a surprise to me.   Frankly, I consider that assumption a bug... but the project's direction is clear:
* https://github.com/biomejs/biome/issues/10314#issuecomment-4415812206
* https://github.com/biomejs/biome/issues/3524#issuecomment-2256032368

This PR updates the docs to align with the default functionality - files are assumed to be ESM by default when unspecified in `package.json`


ematipico:
This file is auto-generated. You need to modify the rust code in the main repository

alexrudd2:
Thanks, I should have looked harder!  https://github.com/biomejs/biome/pull/11707

PR #4543, opened 2026-09-10
docs: grit range function

dyc3:
## Summary

docs for [#11713](https://github.com/biomejs/biome/pull/11713)

### DIFF
--- a/src/content/docs/en/recipes/gritql-plugins.mdx
+++ b/src/content/docs/en/recipes/gritql-plugins.mdx
@@ -667,6 +667,34 @@ the quotes on both sides.
 
 ## Advanced Patterns
 
+### Flag logging within selected lines
+
+Use [`range()`](/reference/gritql/#matching-source-ranges) to limit a plugin to
+known source positions. This rule reports `console.log()` calls contained
+entirely within lines 2 and 3. The calls before and after that range are skipped.
+If a call spans several lines, the entire call must be inside the selected range
+in order to match.
+
+```grit ins="range"
+`console.log($...)` as $call where {
+    $call <: range(start_line=2, end_line=3),
+    register_diagnostic(
+        span = $call,
+        message = "Remove logging from this region.",
+        severity = "warn"
+    )
+}
+```
+
+```js ins={2-3}
+console.log("before");
+console.log("selected");
+console.log("also selected");
+console.log("after");
+```
+
+[Try this example in the Playground](/playground?tab=syntax&pane=GritQL&code=YwBvAG4AcwBvAGwAZQAuAGwAbwBnACgAIgBiAGUAZgBvAHIAZQAiACkAOwAKAGMAbwBuAHMAbwBsAGUALgBsAG8AZwAoACIAcwBlAGwAZQBjAHQAZQBkACIAKQA7AAoAYwBvAG4AcwBvAGwAZQAuAGwAbwBnACgAIgBhAGwAcwBvACAAcwBlAGwAZQBjAHQAZQBkACIAKQA7AAoAYwBvAG4AcwBvAGwAZQAuAGwAbwBnACgAIgBhAGYAdABlAHIAIgApADsACgA%3D&gritQuery=%60console.log%28%24...%29%60+as+%24call+where+%7B%0A++++%24call+%3C%3A+range%28start_line%3D2%2C+end_line%3D3%29%0A%7D%0A)
+
 ### Combine multiple related rules in one file
 
 You can group **multiple independent rules** into a single `.grit` file using a
--- a/src/content/docs/en/reference/gritql.mdx
+++ b/src/content/docs/en/reference/gritql.mdx
@@ -108,6 +108,74 @@ but it gets quickly more interesting when add other operators in the mix:
 }
 ```
 
+## Matching source ranges
+
+Use the `range()` pattern to filter syntax nodes by their position in the source
+file. For example, this query matches calls contained entirely on line 2:
+
+```grit
+range(start_line=2, end_line=2) as $call where {
+    $call <: call_expression()
+}
+```
+
+Given this source, it matches only `second()`:
+
+```js showLineNumbers {2}
+first();
+second();
+third();
+```
+
+`range()` accepts four optional named arguments:
+
+| Argument | Constraint |
+| --- | --- |
+| `start_line` | The node must start on or after this line. |
+| `start_column` | If the node starts on `start_line`, it must start at or after this column. Requires `start_line`. |
+| `end_line` | The node must end on or before this line. |
+| `end_column` | If the node ends on `end_line`, its end position must be at or before this column. Requires `end_line`. |
+
+Arguments must be unsigned integer literals from `0` to `4294967295`.
+Source lines and columns start at **1**. Columns count Unicode scalar values,
+so `é` counts as one column, and a tab also counts as one column.
+
+Omitting `start_line` leaves the start unbounded; omitting `end_line` leaves the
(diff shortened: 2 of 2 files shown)


src/content/docs/en/recipes/gritql-plugins.mdx
```diff
@@ -667,6 +667,98 @@ the quotes on both sides.
...
+```js showLineNumbers ins={2-3}
+console.log("before");
+console.log("selected");
+console.log("also selected");
+console.log("after");
+```
+
+Only the calls on lines 2 and 3 produce diagnostics. The calls before and after
+that range are skipped. A multiline call must fit entirely within the selected
+lines to match.
```
ematipico: > A multiline call must fit entirely within the selected lines to match.

I don't know how to interpret this phrase. What's it talking about? 

> Only the calls on lines 2 and 3 produce diagnostics. The calls before and after that range are skipped.

Isn't this already said before the example?

src/content/docs/en/recipes/gritql-plugins.mdx
```diff
@@ -667,6 +667,98 @@ the quotes on both sides.
...
+
+```js showLineNumbers mark='console.log("b")'
+console.log("before");
+console.log("a"); console.log("b"); console.log("c");
+console.log("after");
+```
+
+Only `console.log("b")` produces a diagnostic. It starts at column 19 and its
+end position is column 35, immediately after `)`. The semicolon is outside the
+call expression.
```
ematipico:
```suggestion
Only `console.log("b")` produces a diagnostic. It starts at column 19 and its
end position is column 35, immediately after `)`. The semicolon is outside of the specified range.
```

src/content/docs/en/recipes/gritql-plugins.mdx
```diff
@@ -667,6 +667,98 @@ the quotes on both sides.
...
+
+Lines and columns start at 1. Each column bound requires its corresponding
+line bound, and applies only on that boundary line.
```
ematipico: > Each column bound requires its corresponding line bound, and applies only on that boundary line.

This feels like jargon. I don't understand what it's talking about

src/content/docs/en/recipes/gritql-plugins.mdx
```diff
@@ -667,6 +667,98 @@ the quotes on both sides.
...
+
+[Try this example in the Playground](/playground?tab=syntax&pane=GritQL&code=YwBvAG4AcwBvAGwAZQAuAGwAbwBnACgAIgBiAGUAZgBvAHIAZQAiACkAOwAKAGMAbwBuAHMAbwBsAGUALgBsAG8AZwAoACIAYQAiACkAOwAgAGMAbwBuAHMAbwBsAGUALgBsAG8AZwAoACIAYgAiACkAOwAgAGMAbwBuAHMAbwBsAGUALgBsAG8AZwAoACIAYwAiACkAOwAKAGMAbwBuAHMAbwBsAGUALgBsAG8AZwAoACIAYQBmAHQAZQByACIAKQA7AAoA&gritQuery=%60console.log%28%24...%29%60+as+%24call+where+%7B%0A++++%24call+%3C%3A+range%28start_line%3D2%2C+start_column%3D19%2C+end_line%3D2%2C+end_column%3D35%29%0A%7D%0A)
+
+### Apply a rule from a chosen line onward
+
+Omit `end_line` to leave the end unbounded. For example, this rule allows an
+initial setup log and reports subsequent `console.log()` calls:
```
ematipico: I would say "Use only `start_line`", because that's what's used in the code. Saying "omit `end_line`" also assumes that users know about `end_line`, which is only mentioned in the previous examples. 

Assume examples are **atomic** and self-explanatory.

src/content/docs/en/recipes/gritql-plugins.mdx
```diff
@@ -667,6 +667,98 @@ the quotes on both sides.
...
+console.log("started");
+console.log("finished");
+```
+
+The calls on lines 2 and 3 produce diagnostics. The call on line 1 is skipped.
+To select a prefix of the file instead, use an end bound such as
+`range(end_line=2)`.
+
+These bounds refer to the current source file. Update them when edits move the
+code you intend to check.
```
ematipico: Avoid the word "bounds". Use simpler words: limits, ranges, etc.

ematipico (commented):
Important information is spread among examples:
- "Update them when edits move the code you intend to check." is third example only
- "Lines and columns start at 1." is in the second example only

I suggest a main paragraph that talks about ranges in general, or we redirect to the GritQL docs, so we don't bother explaining the semantics and how they work, and we limit ourselves in small examples.

src/content/docs/en/recipes/gritql-plugins.mdx
dyc3: > Isn't this already said before the example?

Yeah, but IMO it reads worse without it.

dyc3:
> I suggest a main paragraph that talks about ranges in general, or we redirect to the GritQL docs, so we don't bother explaining the semantics and how they work, and we limit ourselves in small examples.

Then should we bother with the recipes examples at all? Because all the necessary info is already in the reference page.

I don't think linking to the gritql docs is going to help, because it contains _less_ information about how it works.

Also, it seems that the use case in the original issue (enforcing a copyright comment) isn't quite covered yet. I was going to replace the recipes with that, but biome's grit doesn't match comments yet.

src/content/docs/en/recipes/gritql-plugins.mdx
ematipico: Then put everything before the example. No need to shard the explanation

ematipico:
Then we can explain the semantics in one place so we don't fragment everything too much. Or maybe we should have just one example.

dyc3:
Yeah, I'm leaning towards one example

src/content/docs/en/recipes/gritql-plugins.mdx
```diff
@@ -667,6 +667,34 @@ the quotes on both sides.
...
+    $call <: range(start_line=2, end_line=3),
+    register_diagnostic(
+        span = $call,
+        message = "Remove logging from this region.",
+        severity = "warn"
+    )
+}
+```
+
+```js showLineNumbers ins={2-3}
```
yanthomasdev: Line numbers are enabled by default for non-`shell` languages, IIRC
```suggestion
```js ins={2-3}
```

src/content/docs/en/reference/gritql.mdx
```diff
@@ -108,6 +108,74 @@ but it gets quickly more interesting when add other operators in the mix:
 }
 ```
 
+## Matching source ranges
```
yanthomasdev: This page is using "Title Case" instead of "Sentence case", so we should use the former to keep it internally consistent. I plan to move it all to sentence case as I rewrite it.
```suggestion
## Matching Source Ranges
```

src/content/docs/en/reference/gritql.mdx
```diff
@@ -108,6 +108,74 @@ but it gets quickly more interesting when add other operators in the mix:
...
+
+```grit
+range(start_line=2, end_line=2) as $call where {
+    $call <: call_expression()
+}
+```
+
+Given this source, it matches only `second()`:
+
+```js showLineNumbers {2}
```
yanthomasdev:
```suggestion
```js {2}
```

yanthomasdev (commented):
Some nits from me

ematipico (approved):

yanthomasdev (approved):
LGTM!

PR #4558, opened 2026-09-17
feat: add playground link

### LINKED ISSUE #4499: feature: test in playground url
On linting rule pages, a link to the playground with only that rule selected would be nice. Currently you need to scroll through a big list to find a specific rule.

Nice to have: perhaps in the little code snippets on a lint rule page: open playground with this piece of code

ematipico:
## Summary

Closes https://github.com/biomejs/website/issues/4499

The proposal was limited to lint rules, but we could do the same for assist actions.

- Added a `Try in the playground` link at the bottom of each rule and action page. The link is generated using the first invalid example we see during code generation
- The link opens up a new browser window
- The playground now shows a dropdown for the assist actions. Before, the actions were incorrectly placed inside the lint rules dropdown
- The playground now has two hash parameters: `lintRules` and `assistActions`

> Used a coding agent for this small refactor


dyc3 (approved):
Image

nit: The icon is a little off center vertically.

Other than that, works great.

ematipico:
Yeah, I know. I left it like this on purpose. I'll follow up with another PR. This one is big already.

